PR #19, opened 2022-01-08
fix(alonzo): Contemplate tx aux data with multiple plutus scripts

scarmuega:


scarmuega:
related to https://github.com/txpipe/oura/issues/37

PR #23, opened 2022-01-17
Offer a faster way to compute the hash of an encodable object

NicolasDP:
I also bumped to the latest version of cryptoxide. There's been a lot of performance improvements that have been made since then and this would be a shame to not leverage this (though it's mostly for `Curve25519`).

This PR remove the temporary (useless?) serialisation of the `Encode`-able object.  That way we now have a memory overhead allocation to do the hash.

- [ ] there is no way this can fail [`minicbor::Encoder`](https://gitlab.com/twittner/minicbor/-/blob/develop/minicbor/src/encode/encoder.rs#L247-250) have only the `Write` error. In the case of `Blake2b` being the writeable object, there is not error possible (none we can/should handle in the code. Shall we remove the `Result<(), Box<dyn Error>>` ?
- [ ] shall we have a pallas-crypto with the different crypto utils we will be using in our code?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,7 @@ members = [
     "pallas-chainsync",
     "pallas-txsubmission",
     "pallas-localstate",
+    "pallas-crypto",
     "pallas-alonzo",
     "pallas",
 ]
--- a/pallas-alonzo/Cargo.toml
+++ b/pallas-alonzo/Cargo.toml
@@ -12,13 +12,9 @@ authors = [
     "Santiago Carmuega <santiago@carmuega.me>"
 ]
 
-[features]
-crypto = ["cryptoxide"]
-
 [dependencies]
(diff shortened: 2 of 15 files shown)


scarmuega (approved):

scarmuega:
my take on the above:
- allocating memory for the hash was a quick and dirty solution, this is a much better approach.
- yes, we should remove the Result wrapper now that the function is no longer fallible.
- having a pallas-crypto would be great. Would you mind doing the first pass of the module with what you have in mind?

NicolasDP:
@scarmuega let me know if this is the right structure

NicolasDP:
I have kept the `pallas_alonzo::crypto` module but I remove the `crypto` feature. It is obviously used everywhere already.

scarmuega:
@NicolasDP looks great! feel free to merge at your discretion.

scarmuega (approved):

PR #33, opened 2022-01-28
Quick and dirty test of wasm compilation

scarmuega:


scarmuega:
@NicolasDP after your suggestion, I did some tests and managed to do a quick wasm compilation of the alonzo lib without need full no-std support. Thanks!

I haven't figured out yet how to interact with the crypto lib in this context.

- Do you think it's possible to have a wasm-compatible crypto lib? Do you anticipate a reasonable LoE?
- Would it make sense to have the crypto under a feature flag and skip it for wasm?
- Would it make sense to have dumb "hash" primitives somewhere else and make both alonzo lib and crypto lib reference those?

NicolasDP:
you should be able to use `pallas-crypto` for `wasm` target. It is already used by `cardano-serialisation-lib` by Emurgo and `catalyst mobile voting app` by IOG.

PR #35, opened 2022-01-30
fix(alonzo): ExUnits steps overflow

rinor:
fix https://github.com/txpipe/oura/issues/113
based on https://github.com/txpipe/oura/issues/113#issuecomment-1025171447

PR #38, opened 2022-01-31
test: Overflow error in ExUnits

scarmuega:
Adds one more block to the alonzo testing dataset to include an example of a block with a large ExUnits value (issue fixed by @rinor in #35)


mark-stopka (approved):
LGTM

PR #40, opened 2022-02-05
chore: Merge mini-protocols into single crate

### LINKED ISSUE #29: Merge mini-protocols crates into a single one
Each mini-protocol implementation currently lives in its own crate. The rationale was that having this isolation would facilitate development and maintenance.

Now that most of the protocols have gone through at least one development iteration, this granular approach is proving harder to maintain than the alternative. The amount of code for each protocols is low, so there's no sensible compile-time improvement on having independent modules. Managing independent versions is prone to error.

The proposal is to merge the corresponding crates into a new one called `pallas-network` and deprecate the old ones.

Some decisiones are still not clear to me:
- should the `pallas-multiplexer` crate remain independent? could it be used outside of the scope of the Ouroboros mini-protocols?
- when we reach the point of having to implement P2P, would that logic require its own crate given its complexity?

cc @NicolasDP

scarmuega:
closes #29

The goal is to have the following network-related crates:

- pallas-multiplexer
- pallas-miniprotocols

The argument against merging the two crates into a potential "pallas-network" is that it would be an artificial coupling not really required at the code level. Multiplexing is agnostic of the mini-protocols (even agnostic of the cbor encoding). Having it as different crates would allow easier repurposing of either crate.


mark-stopka:
@pavlix maybe you can have a look, may give some insight when rewriting our Ouroboros networking libs for Tokio....

mark-stopka:
> @pavlix maybe you can have a look, may give some insight when rewriting our Ouroboros networking libs for Tokio....

Actually never mind, all issues seems totally out of scope for the networking part, and are related to either `alonzo`, or `byron` ledger rules CBOR (de)serialization....

PR #44, opened 2022-02-09
feat: Implement block cbor probing

scarmuega:


rvcas (approved):
this is cool. do we need to probe for alonzo as well?

scarmuega:
yes, it would be a nice feature to probe for different hard forks on top of eras.

For the time being, since Alonzo primitives are backward compatible with previous Shelley blocks, we should be able to decode any block just by knowing if it's byron or shelley.

PR #47, opened 2022-02-12
feat(primitives): Improve ergonomics for Byron primitives

scarmuega:


mark-stopka (approved):
LGTM

mark-stopka:
I don't know if it is necessarily something we want to test for, but @NicolasDP told me, that there are some JPEGs encoded in some Byron addresses, so maybe we should find some of those early "NFTs" when we will be adding the next wave of blocks for tests =)

PR #49, opened 2022-02-16
feat: Implement rollback buffer

scarmuega:
This PR introduces a memory buffer to handle chain rollbacks.

This structure is intended to facilitate the process of managing rollbacks in a chain sync process. The goal is to keep `points` in memory until they reach a certain depth (# of confirmations). If a rollback happens, the buffer will try to find the intersection, clear the orphaned points and keep the remaining still in memory. Further forward rolls will accumulate from the intersection.

It works by keeping a `VecDeque` data structure of points, where roll-forward operations accumulate at the end of the deque and retrieving confirmed points means to pop from the front of the deque.

Notice that it works by keeping track of points, not blocks. It is meant to be used as a lightweight index where blocks can then be retrieved from a more suitable memory structure / persistent storage.

### DIFF
--- a/pallas-miniprotocols/src/common.rs
+++ b/pallas-miniprotocols/src/common.rs
@@ -7,7 +7,7 @@ pub const TESTNET_MAGIC: u64 = 1097911063;
 pub const MAINNET_MAGIC: u64 = 764824073;
 
 /// A point within a chain
-#[derive(Clone)]
+#[derive(Clone, Eq, PartialEq)]
 pub struct Point(pub u64, pub Vec<u8>);
 
 impl Debug for Point {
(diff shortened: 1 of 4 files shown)


pallas-miniprotocols/src/common.rs
```diff
@@ -7,7 +7,7 @@ pub const TESTNET_MAGIC: u64 = 1097911063;
 pub const MAINNET_MAGIC: u64 = 764824073;
 
 /// A point within a chain
-#[derive(Clone)]
+#[derive(Clone, Eq, PartialEq)]
 pub struct Point(pub u64, pub Vec<u8>);
```
NicolasDP, line 11: what is the `Vec<u8>` here ? is it a block hash ?

scarmuega: Yes, it is the hash. This struct was there before the introduction of pallas-crypto. We should turn it into a `Hash<>`.

PR #93, opened 2022-04-26
fix(primitives): Fix native scripts before/after type serialization

SmaugPool:
`InvalidBefore` should actually be serialized to (valid) "after", and `InvalidHereafter` to "before".

### DIFF
--- a/pallas-primitives/src/alonzo/json.rs
+++ b/pallas-primitives/src/alonzo/json.rs
@@ -62,8 +62,8 @@ impl ToCanonicalJson for super::NativeScript {
                 let scripts: Vec<_> = k.iter().map(|i| i.to_json()).collect();
                 json!({ "type": "atLeast", "required": n, "scripts" : scripts })
             }
-            super::NativeScript::InvalidBefore(slot) => json!({ "type": "before", "slot": slot }),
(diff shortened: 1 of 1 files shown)


pallas-primitives/src/alonzo/json.rs
```diff
@@ -62,8 +62,8 @@ impl ToCanonicalJson for super::NativeScript {
                 let scripts: Vec<_> = k.iter().map(|i| i.to_json()).collect();
                 json!({ "type": "atLeast", "required": n, "scripts" : scripts })
             }
-            super::NativeScript::InvalidBefore(slot) => json!({ "type": "before", "slot": slot }),
-            super::NativeScript::InvalidHereafter(slot) => json!({"type": "after", "slot": slot }),
+            super::NativeScript::InvalidBefore(slot) => json!({ "type": "after", "slot": slot }),
+            super::NativeScript::InvalidHereafter(slot) => json!({"type": "before", "slot": slot }),
```
rvcas, line 66: wait it seemed correct before? is this one of those situations where it's unintuitive, so like on `InvalidBefore` we want to say `after this slot`?

rvcas: @SmaugPool nice I see. Thank you

PR #94, opened 2022-04-26
fix(primitives): Fix native scripts policy id (add missing tag)

SmaugPool:
Native scripts hashes need the tag 0 to compute the correct policy id, see:

https://github.com/input-output-hk/cardano-ledger/blob/736c9295f5896de28fbcedc6a8fc4d509fb78d10/shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/Scripts.hs#L61-L65

https://github.com/input-output-hk/cardano-ledger/blob/736c9295f5896de28fbcedc6a8fc4d509fb78d10/shelley/chain-and-ledger/executable-spec/src/Shelley/Spec/Ledger/Scripts.hs#L163-L171

We can add a comment in the code if you want.

### DIFF
--- a/pallas-crypto/src/hash/hasher.rs
+++ b/pallas-crypto/src/hash/hasher.rs
@@ -78,6 +78,14 @@ macro_rules! common_hasher {
                 hasher.finalize()
             }
 
+            #[inline]
+            pub fn hash_tagged_cbor(data: &impl minicbor::Encode, tag: u8) -> Hash<{ $size / 8 }> {
+                let mut hasher = Self::new();
+                hasher.input(&[tag]);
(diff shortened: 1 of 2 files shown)


pallas-crypto/src/hash/hasher.rs
```diff
@@ -78,6 +78,17 @@ macro_rules! common_hasher {
                 hasher.finalize()
             }
 
+            #[inline]
+            pub fn hash_tagged_cbor(
+                data: &impl minicbor::Encode,
+                tag: &[u8],
```
SmaugPool, line 84: We could also use a single `u8` if preferred as this will likely cover all use cases for a long time.

scarmuega (approved):

PR #100, opened 2022-04-29
feat: Implement Plutus Data hashing / JSON

scarmuega:
After some source code spelunking, this PR adds:

- a mechanism to hash Plutus Data from wire-cbor in an equivalent fashion as the cardano-cli.
- export a Plutus Data struct into JSON format that can be read by the cardano-cli


rvcas (approved):
I think this looks good. Is this in a spec somewhere or was this from reverse-engineering some Haskell?

scarmuega:
I use "spelunking" as an euphemism for "reverse engineering"

PR #132, opened 2022-06-19
Add Babbage nonce/leader vrf extension

AndrewWestberg:
Please review thoroughly and give feedback. I probably didn't do this the most optimal way.

### DIFF
--- a/pallas-traverse/src/header.rs
+++ b/pallas-traverse/src/header.rs
@@ -1,8 +1,9 @@
 use pallas_codec::minicbor;
-use pallas_crypto::hash::Hash;
+use pallas_crypto::hash::{Hash, Hasher};
 use pallas_primitives::ToHash;
 
 use crate::{Error, MultiEraHeader};
+use crate::Era::Byron;
 
 impl<'b> MultiEraHeader<'b> {
     pub fn decode(tag: u8, subtag: Option<u8>, cbor: &'b [u8]) -> Result<Self, Error> {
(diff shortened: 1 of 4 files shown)


pallas-traverse/src/lib.rs
```diff
@@ -97,7 +97,10 @@ pub enum Error {
     UnknownCbor(String),
 
     #[error("Unknown era tag: {0}")]
-    UnkownEra(u16),
+    UnknownEra(u16),
+
+    #[error("Invalid era for request: {0}")]
+    InvalidEra(u16),
```
scarmuega, line 103: I think we could use the `crate::Era` struct for this particular error since it will be thrown from methods where the Era of the data is strictly known (the `UnknownEra` relies on u16 because it is used for errors during the parsing of the value).

pallas-traverse/src/header.rs
```diff
@@ -45,4 +46,57 @@ impl<'b> MultiEraHeader<'b> {
             MultiEraHeader::Byron(x) => x.to_hash(),
         }
     }
+
+    pub fn leader_vrf_output(&self) -> Result<ByteVec, Error> {
```
scarmuega, line 50: Could we make the fn name just `leader_vrf`? Or do you think that the `output` prefix provides more context?

AndrewWestberg: A VRF has two parts. There's the output and the proof. In this case, we only are extending to get the output part of it. If you call it something other than "output" in your code, we can maybe change it.

scarmuega: makes sense now.

pallas-traverse/src/header.rs
```diff
@@ -45,4 +46,57 @@ impl<'b> MultiEraHeader<'b> {
...
+        match self {
+            MultiEraHeader::EpochBoundary(_) => Err(Error::InvalidEra(0)),
+            MultiEraHeader::AlonzoCompatible(x) => Ok(x.header_body.leader_vrf.0.clone()),
+            MultiEraHeader::Babbage(x) => {
+                let mut leader_tagged_vrf: Vec<u8> = vec![0x4C_u8]; /* "L" */
+                leader_tagged_vrf.extend(&*x.header_body.vrf_result.0);
+                let h32 = Hasher::<256>::hash(&leader_tagged_vrf);
+                Ok(ByteVec::from(h32.to_vec()))
```
scarmuega, line 58: I think we can skip the ByteVec wrapper and just return a Hash<32>. The bytevec is a cbor-specific decoding struct which should remain as hidden as possible from the api surface.

pallas-traverse/src/lib.rs
AndrewWestberg: fixed

pallas-traverse/src/header.rs
AndrewWestberg: Went with a simple `Vec<u8>`.

pallas-traverse/src/lib.rs
AndrewWestberg: fixed

PR #137, opened 2022-06-23
feat: Introduce Addresses crate

scarmuega:
This module contains utilities to decode / encode Cardano addresses from to different formats. The entry point to most of the methods is the `Address` enum, which holds the decoded values of either a Byron, Shelley or Stake address.


rvcas (approved):

NicolasDP (approved):
Looking mighty

PR #150, opened 2022-07-05
Pallas Ouroboros  Tx-Mempool-Monitoring Mini-Protocol 

Sbcdn:
This is the Ouroboros Tx-Mempool-Monitoring Miniprotocol to interact with the local Mempool of a Cardano Node. 
I also added the Protocol Versions 11 and 12 for the handshake protocol as it is needed for the monitoring protocol. 

I request that some else also tests the code. 

A first implementation of the Mempool monitoring Mini-Protocol can be found here: 
https://github.com/Sbcdn/sif


scarmuega:
this is awesome!

PR #160, opened 2022-08-06
Getters for witness fields

SebastienGllmt:
I'm not sure what is the policy for flattening fields for Pallas. Presumably based on the way blocks and txs are handled, the policy is to flatten stuff when it's shared between eras? Based on this I added some getters for the witness types

### DIFF
--- a/pallas-traverse/src/witnesses.rs
+++ b/pallas-traverse/src/witnesses.rs
@@ -1,4 +1,3 @@
-use pallas_codec::utils::MaybeIndefArray;
 use pallas_primitives::{alonzo::{self, VKeyWitness, Redeemer, PlutusData, BootstrapWitness, NativeScript}, babbage::{self, PlutusV2Script}};
 
 use crate::MultiEraWitnesses;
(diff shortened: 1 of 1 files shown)


scarmuega:
@SebastienGllmt  thanks for adding this!

> I'm not sure what is the policy for flattening fields for Pallas. Presumably based on the way blocks and txs are handled, the policy is to flatten stuff when it's shared between eras? Based on this I added some getters for the witness types

Yes, exactly, the idea is to flatten anything that shares a common primitive.

Since we're returning a borrow already, do you think that we can expose it as a slice instead of `MaybeIndefArray`?

I was thinking of something like this:

```rust
pub fn bootstrap_witness(&self) -> Option<&[BootstrapWitness]> {
    match self {
        Self::AlonzoCompatible(x) => x.bootstrap_witness.as_ref().map(|x| x.as_ref()),
        Self::Babbage(x) => x.bootstrap_witness.as_ref().map(|x| x.as_ref()),
        _ => None,
    }
}
```

Would great to expose as little as possible of all this nasty cbor wrappers.

SebastienGllmt:
done

scarmuega (approved):

PR #180, opened 2022-09-01
feat: return all assets at a tx out

rvcas:
* loop through multiasset and concat policy + asset name
* ada unit is lovelace

I can probably create a helper that takes a `&mut assets` to reduce some code. The same double for loop is written 3 times

### DIFF
--- a/pallas-codec/src/utils.rs
+++ b/pallas-codec/src/utils.rs
@@ -1,6 +1,6 @@
 use minicbor::{data::Tag, Decode, Encode};
 use serde::{Deserialize, Serialize};
-use std::ops::Deref;
+use std::{fmt, ops::Deref};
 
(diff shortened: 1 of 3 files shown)


MicroProofs (dismissed):

scarmuega:
Looks great. I agree that we can make it a little bit more DRY on a follow-up PR. The other thing I would revisit is using `lovelace` as a string. Maybe we can use an enum along the lines of:

```rust
enum AssetName {
  Lovelace,
  NativeAsset(policy, name)
}

impl ToString for AssetName {
  fn to_string(&self) -> String {
     match self {
        Self::Lovelace => String::from("lovelace"),
        Self::NativeAsset(p, n) => format!("{p}.{n}"),
     }
  }
}
```

PR #181, opened 2022-09-01
feat: add `to_hex` and `to_bech32` methods for address payment and delegation parts

jmhrpr:
While being a nice to have feature in general, this will also allow us to progress filtering by address payment or delegate part in Scrolls in a cleaner way.

Prefixes from [CIP 5](https://cips.cardano.org/cips/cip5/).


jmhrpr:
It's kind of a shame that `to_bech32` is fallible. If it is just because of the fact the characters in the hrp have to be legal then can we just unwrap as acceptable hrps are hardcoded?

scarmuega (approved):

PR #189, opened 2022-09-12
feat(traverse): Provide access to original Datum hash

scarmuega:


pallas-primitives/src/alonzo/model.rs
```diff
@@ -1179,6 +1179,43 @@ pub struct TransactionWitnessSet {
...
 }
 
+#[derive(Encode, Decode, Debug, PartialEq, Clone)]
+#[cbor(map)]
+pub struct MintedWitnessSet<'b> {
+    #[n(0)]
+    pub vkeywitness: Option<Vec<VKeyWitness>>,
+
+    #[n(1)]
+    pub native_script: Option<Vec<NativeScript>>,
```
jmhrpr, line 1189: Should we use KeepRaw for NativeScript? When we decode a NativeScript into an internal representation we decode both definite and indefinite arrays, but we encode to definite. The hash needs to match the initial bytes so that the script hash matches addresses in the inputs.

jmhrpr (approved):
LGTM

pallas-primitives/src/alonzo/model.rs
scarmuega: yes, there are lot of sub-structures that will need this treatment. Off the top of my head, redeemers and metadata. I would prefer to treat each one on a different PR.

PR #196, opened 2022-10-13
fix: Handle undefined CBOR maps in Plutus data

scarmuega:


jmhrpr (approved):
LGTM - we may also need to handle `MapIndef` and `ArrayIndef` for decoding `Metadatum`

PR #198, opened 2022-10-18
feat: Migrate to dumb agents

scarmuega:
Pallas' original approach to mini-protocols was to define autonomous agents that would start and finish a particular interaction from a set of predetermined arguments. The control flow was kept as a private implementation of each agent. The rationale was that by encapsulating this logic, the consumer wouldn't have to know the exact sequence of messages required to progress the state machine of each mini-protocol.

Although valid, this approach ended up being too rigid and limited.

This PR is the 1st of a series of PRs that migrates the original approach into something more flexible. Instead of encapsulating the control within an agent, we provide a facade on top of the state machine with individual, imperative methods to control the state machine in a more atomic way. In this new approach, the control flow is defined by the consumer. The role of Pallas is to ensure that the business invariants defined by the mini-protocol are respected.


pallas-miniprotocols/src/blockfetch/client.rs
```diff
@@ -0,0 +1,171 @@
...
+
+    pub fn recv_request_range(&mut self) -> Result<(), Error> {
+        match self.recv_message()? {
+            Message::StartBatch => {
+                self.0 = State::Streaming;
+                Ok(())
+            }
+            Message::NoBlocks => {
+                self.0 = State::Idle;
+                Err(Error::NoBlocks)
```
jmhrpr, line 129: Is it correct to error for something which is technically a valid state transition? Maybe `request_range` can Error if it receives a NoBlocks because it's like a higher-level method, but `recv_request_range` is the only "low-level" function we have to progress the state machine. But perhaps this is the best Rust-y way as I'm not sure what the alternative return type would be. Not a big deal really.

pallas-miniprotocols/src/blockfetch/client.rs
```diff
@@ -0,0 +1,171 @@
...
+
+    pub fn request_range(&mut self, range: (Point, Point)) -> Result<(), Error> {
+        self.send_request_range(range)?;
+        self.recv_request_range()
+    }
+
+    pub fn recv_streaming(&mut self) -> Result<Option<Body>, Error> {
+        match self.recv_message()? {
+            Message::Block { body } => {
+                self.0 = State::Streaming;
```
jmhrpr, line 143: Unnecessary as we are already in `State::Streaming`

pallas-miniprotocols/src/blockfetch/codec.rs
```diff
@@ -0,0 +1,72 @@
...
+            Message::StartBatch => {
+                e.array(1)?.u16(2)?;
+                Ok(())
+            }
+            Message::NoBlocks => {
+                e.array(1)?.u16(3)?;
+                Ok(())
+            }
+            Message::Block { body } => {
+                e.array(2)?.u16(4)?;
```
jmhrpr, line 31: Missing a CBOR tag here? We decode one in the corresponding decode method

`msgBlock = [4 , #6.24( bytes . cbor block ) ]`

pallas-miniprotocols/src/handshake/client.rs
```diff
@@ -0,0 +1,137 @@
...
+    #[error("attemted to receive message while agency is ours")]
+    AgencyIsOurs,
+    #[error("attemted to send message while agency is theirs")]
+    AgencyIsTheirs,
+    #[error("inbound message is not valid for current state")]
+    InvalidInbound,
+    #[error("outbound message is not valid for current state")]
+    InvalidOutbound,
+    #[error("no intersection point found")]
+    IntersectionNotFound,
```
jmhrpr, line 19: Dead code

pallas-miniprotocols/src/localstate/client.rs
```diff
@@ -0,0 +1,175 @@
...
+        self.send_message(&msg)?;
+        self.0 = State::Querying;
+
+        Ok(())
+    }
+
+    pub fn recv_querying(&mut self) -> Result<Q::Response, Error> {
+        match self.recv_message()? {
+            Message::Result(x) => {
+                self.0 = State::Idle;
```
jmhrpr, line 162: Think this should be `State::Acquired` unless this was intentional

jmhrpr (approved):
I reviewed every changed file other than:
- examples/*
- pallas-miniprotocols/tests/integration.rs

pallas-miniprotocols/src/blockfetch/client.rs
scarmuega: I went through the same mental process that you did. At first I created a custom enum to handle both results (started and no-blocks) as an Ok() variant, but it seemed too verbose for something that was easily handled by an extra error variant.

I think that in most scenarios, receiving no blocks after requesting a range would result in an error.

I like the idea of using a more strict approach that represents the protocol without any assumptions, but then hiding this into a higher-level abstraction as a different method.

pallas-miniprotocols/src/localstate/client.rs
scarmuega: my bad

pallas-miniprotocols/src/blockfetch/codec.rs
scarmuega: This hasn't changed, but maybe we're compensating in other places. Which decode method are you referring to?

jmhrpr: These are the encode method for the messages, and the corresponding decode method for `Message::Block` decodes a tag. I assume that it hasn't been issue because no one has used outbound server messages

pallas-miniprotocols/src/blockfetch/client.rs
scarmuega: moved the error to `fetch_single` and `fetch_range` wrapper methods.

jmhrpr (approved):
LGTM

PR #212, opened 2022-12-10
Match CBOR encoding of plutus data with the haskell implementation.

### LINKED ISSUE #211: Another canonical CBOR problem.
CBOR lists can be encoded in two ways: definite and indefinite length.

As seen for example [here](https://github.com/txpipe/pallas/blob/03c36f57eee8afa27454f3599cd8b8c9faeaee98/pallas-primitives/src/alonzo/model.rs#L1082) the encoding of `PlutusData` - here `Constr` concretely - uses indefinite length lists.
I've seen you changed the implementation all over the place to use indefinite length lists in the commit `4f9fc1e40d94bfcbee532010738f2e50ad74970e`. The comment says "we use indef array by default to match the approach used by the cardano-cli".  So my question is 
 * In what scenario does cardano-cli do this encoding?
 Is it when we provide plutus data for redeemers/datums in the json format?

 And my problem is
  * haskell's `Codec.Serialise.serialise` (which I'd call "canonical" implementation right?) serialises plutus-data `Constructor` using definite length list. When `PlutusData` value gets serialized as a part of plutus script it results in a different script serialization and then the hash.
  
  So it looks like either 
   - the haskell implementations already encode cbor inconsequentialy
   - there's a default that luckily is used by every "official" (iog haskell) implementation
   
And my hope is that you were wrong with how cardano-cli encodes plutus data and we can revert the implementation to use definite length lists (which suits me).

zmrocze:
Fixes #211. 

I've seen you already decided to fix the problem of ambiguous CBOR encodings with `KeepRaw` and keeping the initial bytes to be used for hashing etc.

But it is only a plus to make encodings here follow the ones in the [haskell implementation](https://github.com/input-output-hk/plutus/blob/9538fc9829426b2ecb0628d352e2d7af96ec8204/plutus-core/plutus-core/src/PlutusCore/Data.hs#L140), so I made the change.

I've looked on `PlutusData` encoding only. The differences to haskell implementation were:
 - empty lists and lists of arguments to a constructor are encoded as a special case as 0-length definite length arrays
 - maps are encoded as definite arrays (!)
 - bytestrings over 64 bytes are encoded as indefinite bytestrings
 
 I modified those 3 to match haskell. There is a new type `PlutusBytes`, same as `Bytes` but the encoding/decoding different. There is a test case `plutus_data_isomorphic_decoding_encoding`.

### DIFF
--- a/pallas-codec/src/utils.rs
+++ b/pallas-codec/src/utils.rs
@@ -786,6 +786,90 @@ impl fmt::Display for Bytes {
     }
 }
 
+/// Defined to encode PlutusData bytestring as it is done in the canonical plutus implementation
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+#[serde(into = "String")]
+#[serde(try_from = "String")]
+pub struct PlutusBytes(Vec<u8>);
+
+impl From<Vec<u8>> for PlutusBytes {
+    fn from(xs: Vec<u8>) -> Self {
+        PlutusBytes(xs)
+    }
+}
+
+impl From<PlutusBytes> for Vec<u8> {
+    fn from(b: PlutusBytes) -> Self {
+        b.0
+    }
+}
+
+impl Deref for PlutusBytes {
+    type Target = Vec<u8>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl TryFrom<String> for PlutusBytes {
+    type Error = hex::FromHexError;
+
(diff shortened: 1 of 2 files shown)


zmrocze:
If you like the change, could you make a new release, so that I can start relying on this?
Really I'd want this update inside yours [uplc](https://github.com/txpipe/aiken) crate, so I would need the help once again to make a release there. 

Either way I greatly appreciate your work here!

scarmuega:
This is amazing! I'm still reviewing the PR, but it's just for me to get familiarized with the code. We'll get this merged shortly.

PR #223, opened 2023-02-04
Add constants for known miniprotocols

Quantumplation:
Now consumers of the crate don't have to memorize what channel number means what


Quantumplation:
Based on this chart from the networking spec:
image

scarmuega (approved):

PR #226, opened 2023-02-05
Implement server-side handshake client

Quantumplation:


pallas-miniprotocols/README.md
```diff
@@ -12,15 +12,15 @@ The following architectural decisions were made for this particular Rust impleme
 
 ## Development Status
```
Quantumplation, line 13: Rendered: https://github.com/txpipe/pallas/blob/26eb03d07b89d991d16e60b4a1b47825369e9fe5/pallas-miniprotocols/README.md

pallas-miniprotocols/src/handshake/README.md
```diff
@@ -0,0 +1,93 @@
+# Handshake
```
Quantumplation, line 1: Rendered: https://github.com/txpipe/pallas/blob/26eb03d07b89d991d16e60b4a1b47825369e9fe5/pallas-miniprotocols/src/handshake/README.md

pallas-miniprotocols/src/handshake/server.rs
```diff
@@ -0,0 +1,114 @@
...
+
+    pub fn send_accept_version(&mut self, version: VersionNumber, extra_params: D) -> Result<(), Error> {
+        let message = Message::Accept(version, extra_params);
+        self.send_message(&message)?;
+        self.0 = State::Done;
+
+        Ok(())
+    }
+
+    pub fn send_refuse(&mut self, reason: RefuseReason) -> Result<(), Error> {
```
Quantumplation, line 99: personally I'd name these methods more naturally, `refuse`, `accept_version`, etc; but other protocols make sending/receiving explicit, so I matched that style.  If you're not married to this, I could rename these methods.

pallas-miniprotocols/src/handshake/client.rs
```diff
@@ -1,27 +1,8 @@
...
-    #[error("outbound message is not valid for current state")]
-    InvalidOutbound,
-
-    #[error("error while sending or receiving data through the channel")]
-    ChannelError(ChannelError),
-}
+use super::{Message, RefuseReason, State, VersionNumber, VersionTable, Error};
 
 #[derive(Debug)]
 pub enum Confirmation<D> {
```
Quantumplation, line 8: One thing I noticed is this `Confirmation` type uses the term `Reject` while everything else uses `Refuse`; do we want to make it consistent?

pallas-miniprotocols/src/handshake/protocol.rs
```diff
@@ -1,6 +1,26 @@
 use itertools::Itertools;
```
Quantumplation, line 1: oop, there's a todo in the Message decode that I need to sort out

pallas-miniprotocols/src/handshake/server.rs
scarmuega: the `send_` & `receive_` prefixes are meant for low level functions that only care about sending / receiving a particular message, without any other business logic. Other higher-level functions would orchestrate those low-level send / receive mechanisms.

I like `refuse` and `accept` and since they convey a high-level operation, makes sense to expose that in the api surface. Feel free to change it.

scarmuega (approved):

PR #229, opened 2023-02-14
Make TxBody in TxSubmission generic, removing hard-coded encoding/decoding.

Quantumplation:
We were still baking in some assumptions about encoding/decoding the TxBody, so this makes it generic in the same way as TxId.

We also provide a default implementation for Cardano specific use cases, since that will be 99% of the cases.

### DIFF
--- a/pallas-miniprotocols/src/txsubmission/codec.rs
+++ b/pallas-miniprotocols/src/txsubmission/codec.rs
@@ -2,7 +2,7 @@ use pallas_codec::minicbor::{decode, encode, Decode, Decoder, Encode, Encoder};
 
 use super::{
     protocol::{Message, TxIdAndSize},
-    EraTxId, TxBody,
+    EraTxId, EraTxBody,
 };
 
 impl<TxId: Encode<()>> Encode<()> for TxIdAndSize<TxId> {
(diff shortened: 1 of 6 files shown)


pallas-miniprotocols/src/txsubmission/codec.rs
```diff
@@ -84,18 +84,30 @@ impl<TxId: Encode<()>> Encode<()> for Message<TxId> {
...
+impl<'b> Decode<'b, ()> for EraTxBody {
     fn decode(d: &mut Decoder<'b>, _ctx: &mut ()) -> Result<Self, decode::Error> {
         d.array()?;
-        // TODO: the TxBody encoding here needs to be pinned down and parameterized, the
-        // same way we did TxId!
-        d.u16()?; // Era?
-        d.tag()?; // tag 24?
-        Ok(TxBody(d.bytes()?.to_vec()))
+        let era = d.u16()?;
+        let tag = d.tag()?; // tag 24? TODO(pi): Not sure what the 24 here means
```
Quantumplation, line 91: oh, 24 means "CBOR bytestring", i can probably remove this comment

scarmuega (approved):

PR #238, opened 2023-03-23
Add clone impls to some errors

cameron1024:
Very small PR that adds `Clone` impls to a couple of error types. I have some error types in my own crate that have `Clone` bounds, and IMO it's better for the underlying error to implement `Clone`, rather than chucking it in an `Arc`.

Thanks :grin:

### DIFF
--- a/pallas-miniprotocols/src/blockfetch/client.rs
+++ b/pallas-miniprotocols/src/blockfetch/client.rs
@@ -6,7 +6,7 @@ use crate::common::Point;
 
 use super::{Message, State};
 
-#[derive(Error, Debug)]
+#[derive(Error, Debug, Clone)]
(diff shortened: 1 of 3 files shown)


scarmuega:
Hi @cameron1024 — thanks for the contribution, and apologies for the very long delay in getting back to you.

Since this PR was opened, the repository has been restructured: `pallas-miniprotocols` and `pallas-multiplexer` no longer exist (their functionality was folded into `pallas-network`), so this patch can no longer be rebased as-is. The error types it targeted have been reorganized as well.

If you still need `Clone` on the current network/error types, we'd be happy to look at a fresh PR against `main`. Closing this one for housekeeping — thanks again for taking the time.

PR #248, opened 2023-04-12
Hydra transaction with modified ledger params fails to deserialise.

matiwinnetou:
Here we can see a hydra transaction with modified ledger params.

```
minFeeA: 1
minFeeB: 1
utxoCostPerWord: 0,
utxoCostPerByte: 0
maxTxExecutionUnits": {
        "memory": 9223372036854775806,
        "steps": 9223372036854775806
},
```

Unfortunately this fails to be deserialised, all attempts to fix this for the moment failed.


matiwinnetou:
```
---- block::tests::test_hydra_tx_with_adjusted_ledger_params stdout ----
thread 'block::tests::test_hydra_tx_with_adjusted_ledger_params' panicked at 'called `Result::unwrap()` on an `Err` value: Error { err: Message, pos: None, msg: "unknown cbor data type:i32 for Alonzo Value enum." }', pallas-traverse/src/block.rs:225:306
stack backtrace:
```

PR #263, opened 2023-06-21
Support windows

OlofBlomqvist:
Just some changes to make it possible to clone and build the repo on Windows.

- Rename aux.rs to auxilary.rs since aux is a reserved name on Windows.
- Place unix socket code behind platform flag.


scarmuega:
🙏  thanks!

scarmuega (approved):

PR #271, opened 2023-07-23
WIP: Transaction Builder

cfcosta:


pallas-txbuilder/src/transaction/mod.rs
```diff
@@ -0,0 +1,50 @@
...
+    pub fn build(self) -> TransactionInput {
+        TransactionInput {
+            transaction_id: self.transaction_id,
+            index: self.index,
+        }
+    }
+}
+
+#[derive(Debug, Encode, Decode, Clone)]
+pub struct Transaction {
```
rvcas, line 35: Was it a pain or not reasonable to reuse the one defined in primitives?

cfcosta: I'm going to check again, but afaik there's only `TransactionBody` in primitives, not `Transaction` itself.

rvcas: Oh right that's correct

pallas-txbuilder/src/lib.rs
```diff
@@ -0,0 +1,59 @@
...
+mod builder;
+mod fee;
+mod transaction;
+
+pub mod prelude;
+
+#[derive(Debug, Clone)]
+pub struct NetworkParams {
+    pub genesis_values: GenesisValues,
+    pub min_utxo_value: u64,
```
caike, line 12: Heads up I believe `min_utxo_value` is no longer involved in calculating tx fees (and no longer present in the protocol params at all). It's been replaced by [utxoCostPerByte](https://github.com/input-output-hk/cardano-node/blob/5006b56fc73a43a5363b5a380af0c98b86af4589/bench/tx-generator/data/protocol-parameters-v8.json#L384) and [the calculation is now based on the size of the CBOR.](https://cips.cardano.org/cips/cip55/#thenewminimumlovelacecalculation)

cfcosta: Oh, I'm not using it for fee calculation, but to validate output creation.

pallas-txbuilder/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+license = "Apache-2.0"
+readme = "README.md"
+authors = [
+  "Santiago Carmuega <santiago@carmuega.me>",
+  "Cainã Costa <me@cfcosta.com>"
+]
+
+[dependencies]
+hex = "0.4.3"
+minicbor = { version = "0.19.1", features = ["minicbor-derive"] }
```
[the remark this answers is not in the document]
cfcosta: Agree, let me do that.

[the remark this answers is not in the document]
cfcosta: Actually, it didn't work :(
What happens is: because of the way that the proc macros work, if you define `Encode` or `Decode` for a type, you need to have a `::minicbor` available, or it doesn't work. So whether or not I use it, I need to have `minicbor` on my Cargo.toml :(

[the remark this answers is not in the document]
rvcas: I see

pallas-txbuilder/src/asset.rs
```diff
@@ -0,0 +1,54 @@
...
+#[derive(Debug, Clone, PartialEq, Eq, thiserror::Error)]
+pub enum AssetError {
+    #[error("Invalid asset name")]
+    InvalidAssetName(String),
+}
+
+#[derive(Debug, Clone, Default)]
+pub struct MultiAsset<T> {
+    assets: IndexMap<PolicyId, Vec<(Bytes, T)>>,
+    _marker: PhantomData<T>,
```
MitchTurner, line 16: You can probably remove the marker, right? Since `assets` will have a concrete `T` at compilation.

pallas-txbuilder/src/asset.rs
```diff
@@ -0,0 +1,54 @@
...
+use pallas_primitives::babbage::PolicyId;
+
+#[derive(Debug, Clone, PartialEq, Eq, thiserror::Error)]
+pub enum AssetError {
+    #[error("Invalid asset name")]
+    InvalidAssetName(String),
+}
+
+#[derive(Debug, Clone, Default)]
+pub struct MultiAsset<T> {
```
MitchTurner, line 14: Maybe rename this `MultiAssetBuilder` just not to collide with `pallas_primitives::babbage::Multiasset<T>`

pallas-txbuilder/src/builder.rs
```diff
@@ -0,0 +1,201 @@
...
+    network_params: NetworkParams,
+    mint: Option<MultiAsset<i64>>,
+    required_signers: Vec<AddrKeyhash>,
+    valid_after: Option<u64>,
+    valid_until: Option<u64>,
+    certificates: Vec<Certificate>,
+    plutus_data: Vec<PlutusData>,
+}
+
+impl TransactionBuilder {
```
MitchTurner, line 33: Are you considering doing any sort of automatic input selection? Or are you saying that's out-of-scope. I can get behind either answer, I just know CML has that.

pallas-txbuilder/src/builder.rs
```diff
@@ -0,0 +1,201 @@
...
+
+    pub fn valid_after(mut self, timestamp: u64) -> Self {
+        self.valid_after = Some(timestamp);
+        self
+    }
+
+    pub fn valid_until(mut self, timestamp: u64) -> Self {
+        self.valid_until = Some(timestamp);
+        self
+    }
```
MitchTurner, line 96: Thank you.

pallas-txbuilder/src/builder.rs
cfcosta: Out of scope for now, I plan on adding it on a more advanced implementation later, if there's demand for it.

pallas-txbuilder/src/asset.rs
cfcosta: True! Will do that.

pallas-txbuilder/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+
+    pub fn network_id(&self) -> u64 {
+        self.genesis_values.network_id
+    }
+
+    pub fn timestamp_to_slot(&self, timestamp: u64) -> Option<u64> {
+        (timestamp / self.genesis_values.shelley_slot_length as u64)
+            .checked_sub(self.genesis_values.shelley_known_time)
+    }
+}
```
MitchTurner, line 37: Shouldn't this be something more like
```
(timestamp - shelley_time) / slot_length
```
(or the checked version of that)?

This is subtracting a time from a slot I think.

pallas-txbuilder/src/asset.rs
cfcosta: Fixed!

pallas-txbuilder/src/lib.rs
cfcosta: True! Fixed.

pallas-txbuilder/src/transaction/mod.rs
rvcas: actually turns out it's right here

https://github.com/cfcosta/pallas/blob/ea9c4e8b59fe5cbaaf73c649b44347a066f70960/pallas-primitives/src/babbage/model.rs#L699

PR #272, opened 2023-07-24
feat(pallas-traverse): add network id to genesis values

cfcosta:


cfcosta:
@scarmuega this is "necessary" for the transaction builder, but it is a small change that might help other people too.

scarmuega:
nice!

PR #279, opened 2023-08-13
feat: Windows Socket Connections

OlofBlomqvist:
This PR adds a platform specific connect method to the NodeClient so that it can connect to cardano nodes via named pipes on Windows.

Client setup as usual:
```rust
let mut client = 
  pallas_network::facades::NodeClient::connect(
      r#"\\.\pipe\cardano-node-preprod"#, 
      self.configuration.magic
  ).await
```
Run node with named pipe:
```powershell
cardano-node.exe run ....... --socket-path "\\.\pipe\cardano-node-preprod"
```

### DIFF
--- a/pallas-network/src/facades.rs
+++ b/pallas-network/src/facades.rs
@@ -8,7 +8,13 @@ use tracing::{debug, error};
 #[cfg(not(target_os = "windows"))]
 use tokio::net::UnixListener;
 
-use crate::miniprotocols::handshake::{n2c, n2n, Confirmation, VersionNumber};
+use crate::miniprotocols::handshake::{
+    n2c::VersionData, 
+    n2n, VersionTable,
+    Confirmation, 
+    VersionNumber
+};
+
 use crate::miniprotocols::PROTOCOL_N2N_HANDSHAKE;
 use crate::{
     miniprotocols::{
@@ -154,13 +160,8 @@ pub struct NodeClient {
 }
 
 impl NodeClient {
-    #[cfg(not(target_os = "windows"))]
-    pub async fn connect(path: impl AsRef<Path>, magic: u64) -> Result<Self, Error> {
-        debug!("connecting");
-
-        let bearer = Bearer::connect_unix(path)
-            .await
-            .map_err(Error::ConnectFailure)?;
+    
+    async fn connect_bearer(bearer:Bearer,versions: VersionTable<VersionData>) -> Result<Self, Error> {
 
         let mut plexer = multiplexer::Plexer::new(bearer);
 
@@ -170,7 +171,6 @@ impl NodeClient {
 
         let plexer_handle = tokio::spawn(async move { plexer.run().await });
 
-        let versions = handshake::n2c::VersionTable::v10_and_above(magic);
         let mut client = handshake::Client::new(hs_channel);
 
         let handshake = client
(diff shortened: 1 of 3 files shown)


pallas-network/src/multiplexer.rs
```diff
@@ -63,6 +67,7 @@ pub struct Segment {
 #[cfg(target_os = "windows")]
 pub enum Bearer {
     Tcp(TcpStream),
+    Socket(NamedPipeClient)
```
scarmuega, line 70: could we use `NamedPipe` for the enum member instead of `Socket`?
because I'm afraid that `Socket` can be misinterpreted

pallas-network/src/multiplexer.rs
```diff
@@ -84,25 +89,45 @@ impl Bearer {
         Ok((Self::Tcp(stream), addr))
     }
 
+    #[cfg(windows)]
+    pub async fn connect_pipe(pipe_name: impl AsRef<std::ffi::OsStr>) -> Result<Self, tokio::io::Error>{
```
scarmuega, line 93: could we rename the fn as `connect_named_pipe`?

OlofBlomqvist: code is updated to "connect_named_pipe" but im still seeing it as "connect_pipe" in the comment here for some reason

scarmuega (dismissed):

scarmuega:
@OlofBlomqvist looks good! if you don't mind fixing the merge conflict on your side, I think we're good to go.

OlofBlomqvist:
@scarmuega I disabled the new server side stuff on windows now - wouldn't build otherwise. Can probably look in to adding windows support for that using named pipes as well at some point though

pallas-network/src/facades.rs
```diff
@@ -191,6 +191,43 @@ impl NodeClient {
...
+            error!(?reason, "handshake refused");
+            return Err(Error::IncompatibleVersion);
+        }
+
+        Ok(Self {
+            plexer_handle,
+            handshake,
+            chainsync: chainsync::Client::new(cs_channel),
+            statequery: localstate::Client::new(sq_channel),
+        })
```
scarmuega, line 228: can we move this block into it's own platform-agnostic private function called `connect_bearer(bearer, magic)`?

Just to avoid repetition of logic that is shared between platforms.

pallas-network/src/multiplexer.rs
```diff
@@ -92,25 +98,45 @@ impl Bearer {
...
             Bearer::Unix(x) => x.readable().await,
+
+            #[cfg(target_os = "windows")]
+            Bearer::NamedPipe(x) => x.readable().await
+            
         }
     }
 
-    fn try_read(&mut self, buf: &mut [u8]) -> tokio::io::Result<usize> {
+    async fn try_read(&mut self, buf: &mut [u8]) -> tokio::io::Result<usize> {
```
scarmuega, line 133: why is the fn signature changing to `async`? is this a requirement for named pipes?

OlofBlomqvist: nope, not needed - reverting it

pallas-network/src/facades.rs
OlofBlomqvist: Created now but with this sig:
```
async fn connect_bearer(bearer:Bearer,versions: VersionTable<VersionData>) -> Result<Self, Error>  {...}
```

I am not sure if it is due to a bug or working as designed , but it is only possible to connect to the pipe using only_v10 - otherwise we get back this type of error: 

`ERROR pallas_network::facades: handshake refused, reason: HandshakeDecodeError(32783, "unknown encoding: TList [TInt 1,TBool True]")`

scarmuega (approved):

PR #282, opened 2023-08-25
feat: server facades 

jmhrpr:


pallas-network/src/facades.rs
```diff
@@ -81,6 +85,63 @@ impl PeerClient {
...
+
+        if let Some(ver) = accepted_version {
+            Ok(Self {
+                plexer_handle,
+                version: ver,
+                chainsync: server_cs,
+                blockfetch: server_bf,
+            })
+        } else {
+            Err(Error::IncompatibleVersion)
```
scarmuega, line 127: I think that we should explicitly abort the plexer async thread if we don't find a compatible version

jmhrpr: So call `self.abort()` before returning the error?

scarmuega (approved):

PR #283, opened 2023-09-06
chore: fix lint issues

scarmuega:


scarmuega:
@jmhrpr notice that I renamed `Error` structs inside mini-protocols to `ClientError` and `ServerError` because clippy discovered that we were shadowing them in the parent re-export.

jmhrpr (approved):

PR #289, opened 2023-09-13
LocalTxSubmission client

oskin1:


scarmuega (dismissed):

oskin1:
FYI, I couldn't make it to work with my local cardano-node 8.1.2.
It fails with "ERROR pallas_network::multiplexer: empty bearer" after SubmitTx is sent and response from the node is being awaited.

oskin1:
Finally made it to work with the node. It turned out I was using wrong EraId prefix on my side when testing. Also fixed RejectReason decoder.

scarmuega (approved):

scarmuega:
amazing!

PR #298, opened 2023-10-03
fix(network): fix bad codec for tx monitoring replies

scarmuega:


pallas-network/src/miniprotocols/txmonitor/codec.rs
```diff
@@ -80,12 +80,13 @@ impl<'b> Decode<'b, ()> for Message {
             // find the specs
             4 => Ok(Message::AwaitAcquire),
             5 => Ok(Message::RequestNextTx),
-            6 => match d.array()? {
-                Some(_) => {
-                    let cbor: pallas_codec::utils::CborWrap<Tx> = d.decode()?;
-                    Ok(Message::ResponseNextTx(Some(cbor.unwrap())))
+            6 => match d.datatype()? {
```
teebaumcrypto, line 83: First, client must always acquire a snapshot and hold onto it for subsequent queries. To list all queries, one must call nextTransaction repeatedly until it yields null
(from ogmios)

therefore with the try on datatype, it fails down the line. there is the possible that the node answers with null.

PR #305, opened 2023-10-05
network: Fix return of empty value in the ResponseNextTx

teebaumcrypto:
Rationale:
the node answers with [129, 6] if it sent all tx's from the snapshot

More information:
https://ogmios.dev/mini-protocols/local-tx-monitor/#how-to-use
```
1. acquireMempool  → acquireMempoolResponse
2. nextTransaction → nextTransactionResponse t0
3. nextTransaction → nextTransactionResponse t1
4. nextTransaction → nextTransactionResponse t2
5. nextTransaction → nextTransactionResponse null
```

### DIFF
--- a/pallas-network/src/miniprotocols/txmonitor/codec.rs
+++ b/pallas-network/src/miniprotocols/txmonitor/codec.rs
@@ -80,14 +80,19 @@ impl<'b> Decode<'b, ()> for Message {
             // find the specs
             4 => Ok(Message::AwaitAcquire),
             5 => Ok(Message::RequestNextTx),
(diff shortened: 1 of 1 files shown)


scarmuega:
@teebaumcrypto makes sense, good catch!

Do you have a CBOR snippet that we can use a a decoding test? We have one for a response that includes a tx, but not one for an empty response.

teebaumcrypto:
@scarmuega I don't have a CBOR for testing, because the empty response just returns [129, 6].
I have added a test case for this. (I'm not sure how to assert it correctly

scarmuega (approved):

PR #306, opened 2023-10-05
network: partial fix local tx submission, move RejectReason out of Error

teebaumcrypto:
The erorr should not mix a rust error with a protocol error.


teebaumcrypto:
cbor is skipped if it's a rejection, i think decoder of rejectionreason somehow doesn't return it correctly to the decoder, that it only removes an array and the label from the self.temp value, rather than the whole cbor.
this "hack" fixes it. so the self.temp value gets cleared in the channel.

teebaumcrypto:
if the node rejects a tx with an error from a plutus script, it sends directly the reason, therefore if the first element isn't an array, we expect it to be a rejection.

scarmuega (approved):

PR #309, opened 2023-10-09
Added tx monitor example

kalomaaan:


examples/n2c-miniprotocols/src/main.rs
```diff
@@ -42,20 +42,54 @@ async fn do_chainsync(client: &mut NodeClient) {
...
-            .with_max_level(tracing::Level::TRACE)
+            .with_max_level(tracing::Level::INFO)
             .finish(),
     )
     .unwrap();
 
     // we connect to the unix socket of the local node. Make sure you have the right
     // path for your environment
-    let socket_path = "/tmp/node.socket";
+    let socket_path = "/home/kalomaaantokyo/cardano-my-node/db/socket";
```
rvcas, line 92: Perhaps just read this from an env var?

examples/n2c-miniprotocols/src/main.rs
```diff
@@ -68,14 +102,17 @@ async fn main() {
         .unwrap();
 
     // execute an arbitrary "Local State" query against the node
-    do_localstate_query(&mut client).await;
+    //do_localstate_query(&mut client).await;
 
     // execute the chainsync flow from an arbitrary point in the chain
-    do_chainsync(&mut client).await;
+    //do_chainsync(&mut client).await;
```
scarmuega, line 108: we should maintain the previous examples

examples/n2c-miniprotocols/src/main.rs
scarmuega: @kalomaaan an env var as @rvcas suggests would be great, but if not possible in this PR, lets keep the previous convention for the socket path

PR #314, opened 2023-10-22
Fix LocalTxSubmission node response decoder

oskin1:
The fix is extracted from #306 since in the original PR the fix is mixed with other breaking changes.
This fixes `ERROR pallas_network::multiplexer: err=Error { err: TypeMismatch(Array), pos: Some(1), msg: "expected u16" }` for me.


teebaumcrypto:
The rational behind moving the Reject Reason out of the Error is because the Error is thrown when the channel should be aborted and not be used again.
A Reject of a tx doesn't make the channel unusable, so therefore I recommend making this "breaking" change.

The return Ok is either Accepted or Rejected, but both are valid and should also be treated as "Ok" value, rather than an Error.

oskin1:
@teebaumcrypto Yes, I completely agree that changes in return type signatures introduced in the original PR better reflect the semantics. 
I just wanted to separate decoder fix from breaking changes with a hope that it will be merged quicker.

PR #323, opened 2023-11-01
pallas-primitives: fix PostAlonzoAuxiliaryData

rvcas:
I noticed a small difference between the cddl and the rust definition for `PostAlonzoAuziliaryData`.

I wasn't sure if it was intentional or a mistake but I made the adjustment and all tests still pass.

If this is intentional feel free to close this PR. It was a simple adjustment so no worries.

### DIFF
--- a/pallas-primitives/src/alonzo/model.rs
+++ b/pallas-primitives/src/alonzo/model.rs
@@ -1333,7 +1333,10 @@ pub struct PostAlonzoAuxiliaryData {
     pub native_scripts: Option<Vec<NativeScript>>,
 
     #[n(2)]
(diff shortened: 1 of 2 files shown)


scarmuega:
I don't think we should change this in Alonzo primitives. The plutus_v1 vs v2 was introduced in Babbage. The babbage primitives already contemplate that CDDL change: https://github.com/txpipe/pallas/blob/2b5f22def2a76341337417533b7116e13ac9880c/pallas-primitives/src/babbage/model.rs#L503

rvcas:
Yep, my bad. I must have gotten confused late at night. Closing now.

PR #331, opened 2023-11-13
Fix localtxmonitor decoder

oskin1:


oskin1:
Extracted from #309 so that the fix can be merged separately of non-related things.

PR #335, opened 2023-11-17
add epoch number struct and encode/decode implementations for `hardforkquery` and `ledgerquery`

falcucci:
those were missing to make the localstate query test fixtures.

for now supporting the block epoch number.

### DIFF
--- a/pallas-network/src/miniprotocols/localstate/queries_v16/mod.rs
+++ b/pallas-network/src/miniprotocols/localstate/queries_v16/mod.rs
@@ -80,6 +80,9 @@ pub struct SystemStart {
     pub picoseconds_of_day: u64,
 }
 
+#[derive(Debug, Encode, Decode, PartialEq)]
+pub struct EpochNo(#[n(0)] pub u32);
+
 pub async fn get_chain_point(client: &mut Client) -> Result<Point, ClientError> {
     let query = Request::GetChainPoint;
     let result = client.query(query).await?;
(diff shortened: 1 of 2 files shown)


pallas-network/src/miniprotocols/localstate/queries_v16/mod.rs
```diff
@@ -80,6 +80,9 @@ pub struct SystemStart {
     pub picoseconds_of_day: u64,
 }
 
+#[derive(Debug, Encode, Decode, PartialEq)]
+pub struct EpochNo(#[n(0)] pub u32);
+
```
scarmuega, line 85: why do we need this wrapper?
I would prefer to use the integer directly, unless there's some benefit I'm not seeing.

falcucci: this is for test only purpose to send cbor results and assertions more clear.

e.g

```rust
// send results
let result = AnyCbor::from_encode(EpochNo(8));


// asserts
assert_eq!(epoch, Epoch(8));
```

instead of passing 

```rust
// send results
let result = AnyCbor::from_encode([8]);

// asserts
assert_eq!(epoch, [8];
```

falcucci: done @scarmuega

scarmuega (approved):

PR #338, opened 2023-11-20
Transaction Builder

jmhrpr:


pallas-txbuilder/src/transaction/model.rs
```diff
@@ -0,0 +1,712 @@
...
+    pub datums: Option<HashMap<DatumHash, DatumBytes>>,
+    pub redeemers: Option<Redeemers>,
+    pub script_data_hash: Option<Bytes32>,
+    pub signature_amount_override: Option<u8>,
+    pub change_address: Option<Address>,
+    // pub certificates: TODO
+    // pub withdrawals: TODO
+    // pub updates: TODO
+    // pub auxiliary_data: TODO
+    // pub phase_2_valid: TODO
```
rvcas, line 45: Oh this is exciting, let's start moving Aiken's uplc crate to pallas-plutus when we start building this part? 

cc @KtorZ

pallas-txbuilder/src/transaction/model.rs
```diff
@@ -0,0 +1,712 @@
...
+#[derive(PartialEq, Eq, Hash, Debug)]
+pub enum RedeemerPurpose {
+    Spend(Input),
+    Mint(PolicyId),
+    // Reward TODO
+    // Cert TODO
+}
+
+#[derive(Serialize, Deserialize, PartialEq, Eq, Debug)]
+pub struct ExUnits {
```
rvcas, line 559: potential candidate for `pallas-plutus`? We have the same type in the Aiken uplc crate so we can use the same one in both situations

pallas-txbuilder/src/transaction/model.rs
```diff
@@ -0,0 +1,712 @@
...
+                let mut vkey_witnesses = tx.transaction_witness_set.vkeywitness.unwrap_or_default();
+
+                vkey_witnesses.push(babbage::VKeyWitness {
+                    vkey: Vec::from(pubkey.as_ref()).into(),
+                    signature: Vec::from(signature.as_ref()).into(),
+                });
+
+                tx.transaction_witness_set.vkeywitness = Some(vkey_witnesses);
+
+                self.tx_bytes = tx.encode_fragment().unwrap().into();
```
rvcas, line 649: Doesn't this mean the hash field should be updated as well?

jmhrpr: Transaction hash is only of the transaction body, here we only modify the transaction witnesses

rvcas: oh duh, ok

PR #361, opened 2023-12-17
`anycbor` struct: make the `inner` property publicly accessible

falcucci:
A minor nip to help https://github.com/input-output-hk/mithril/pull/1403 to be finished.

### DIFF
--- a/pallas-codec/src/utils.rs
+++ b/pallas-codec/src/utils.rs
@@ -665,7 +665,7 @@ impl<C, T> minicbor::Encode<C> for KeepRaw<'_, T> {
 /// ```
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone)]
 pub struct AnyCbor {
-    inner: Vec<u8>,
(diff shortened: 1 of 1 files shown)


pallas-codec/src/utils.rs
```diff
@@ -665,7 +665,7 @@ impl<C, T> minicbor::Encode<C> for KeepRaw<'_, T> {
 /// ```
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone)]
 pub struct AnyCbor {
-    inner: Vec<u8>,
+    pub inner: Vec<u8>,
```
scarmuega, line 668: why do we need this?
If you want read-only access you can use `raw_bytes()`
if you want to get the vec you can use `unwrap()`

falcucci: the inner is created only when it has a ready to use instance,

in my case I already have the bytes so was doing something like

```
let datum = AnyCbor { inner: bytes };
```

PR #364, opened 2023-12-20
refactor(network): split bearer into read/write

scarmuega:


examples/n2n-miniprotocols/src/main.rs
```diff
@@ -130,55 +139,33 @@ async fn main() {
...
-                            }
-                        }
-                    }
-                    Err(error) => {
-                        tracing::error!("keepalive error: {:?}", error);
-                    }
-                }
-            }
+        let (chainsync_result, keepalive_result) =
+            tokio::try_join!(chainsync_handle, keepalive_handle)
```
AndrewWestberg, line 157: Great improvement!

pallas-network/src/multiplexer.rs
```diff
@@ -63,112 +65,126 @@ pub struct Segment {
...
-    Tcp(TcpStream),
+    Tcp(tcp::TcpStream),
 
     #[cfg(unix)]
-    Unix(UnixStream),
-
-    #[cfg(windows)]
-    NamedPipe(NamedPipeClient),
+    Unix(unix::UnixStream),
+    // #[cfg(windows)]
```
AndrewWestberg, line 72: dropping windows support?

pallas-network/src/multiplexer.rs
```diff
@@ -63,112 +65,126 @@ pub struct Segment {
...
-        // add tcp_keepalive
+    fn configure_tcp(stream: &tcp::TcpStream) -> IOResult<()> {
         let sock_ref = socket2::SockRef::from(&stream);
         let mut tcp_keepalive = socket2::TcpKeepalive::new();
         tcp_keepalive = tcp_keepalive.with_time(tokio::time::Duration::from_secs(20));
         tcp_keepalive = tcp_keepalive.with_interval(tokio::time::Duration::from_secs(20));
-        let _ = sock_ref.set_tcp_keepalive(&tcp_keepalive);
-        // add tcp_nodelay
-        let _ = sock_ref.set_nodelay(true);        
+        sock_ref.set_tcp_keepalive(&tcp_keepalive)?;
```
AndrewWestberg, line 82: Were you going to move the socket2 tcp_keepalive stuff behind a feature flag and the nodelay to set with tokio?  Or was that in a future PR?

pallas-network/src/multiplexer.rs
```diff
@@ -63,112 +65,126 @@ pub struct Segment {
...
-            Ok(Ok(stream)) => Ok(stream),
-            Ok(Err(err)) => Err(err),
-            Err(_) => Err(tokio::io::Error::new(tokio::io::ErrorKind::TimedOut, "connection timed out")),
+    pub async fn connect_tcp_timeout(
+        addr: impl tcp::ToSocketAddrs,
+        timeout: std::time::Duration,
+    ) -> IOResult<Self> {
+        select! {
+            result = Self::connect_tcp(addr) => result,
+            _ = tokio::time::sleep(timeout) => Err(tokio::io::Error::new(tokio::io::ErrorKind::TimedOut, "connect timeout")),
```
AndrewWestberg, line 100: This is a neat concept. Very clean.

pallas-network/src/multiplexer.rs
```diff
@@ -63,112 +65,126 @@ pub struct Segment {
...
-    async fn flush(&mut self) -> tokio::io::Result<()> {
+    async fn flush(&mut self) -> IOResult<()> {
         match self {
-            Bearer::Tcp(x) => x.flush().await,
+            Self::Tcp(x) => x.flush().await,
 
             #[cfg(unix)]
-            Bearer::Unix(x) => x.flush().await,
-
-            #[cfg(windows)]
```
AndrewWestberg, line 170: comment out instead of delete like the other windows stuff?

pallas-network/src/multiplexer.rs
scarmuega: only temporary, I'll add it back once I find a way to test. It will be before next release but in a different branch.

pallas-network/src/multiplexer.rs
scarmuega: future PR, this one is already a big one

pallas-network/src/multiplexer.rs
```diff
@@ -198,85 +214,90 @@ pub enum Error {
...
+);
 
-impl SegmentBuffer {
-    pub fn new(bearer: Bearer) -> Self {
-        Self(bearer, Vec::with_capacity(BUFFER_LEN))
+pub struct Demuxer(BearerReadHalf, Egress);
+
+impl Demuxer {
+    pub fn new(bearer: BearerReadHalf) -> Self {
+        let egress = tokio::sync::broadcast::channel(10_000);
```
AndrewWestberg, line 231: Make `10_000` a constant and increase to `100_000`. On my fast machine with localhost connection, I was able to overflow when it was only 10k.

pallas-network/src/multiplexer.rs
```diff
@@ -198,85 +214,90 @@ pub enum Error {
...
 
-        trace!("draining segment buffer");
-        let segment = self.1.drain(..segment_size);
-        let payload = segment.skip(HEADER_LEN).collect();
+pub struct Muxer(BearerWriteHalf, Clock, Ingress);
 
-        Ok((header.protocol, payload))
+impl Muxer {
+    pub fn new(bearer: BearerWriteHalf) -> Self {
+        let ingress = tokio::sync::mpsc::channel(100); // TODO: define buffer
```
AndrewWestberg, line 292: Make 100 a `const`

pallas-network/src/multiplexer.rs
scarmuega: yes, good catch

pallas-network/src/multiplexer.rs
```diff
@@ -198,85 +214,90 @@ pub enum Error {
...
-        protocol: u16,
-        clock: &Instant,
-        payload: &[u8],
-    ) -> Result<(), std::io::Error> {
+    async fn write_segment(&mut self, protocol: u16, payload: &[u8]) -> Result<(), std::io::Error> {
         let header = Header {
             protocol,
-            timestamp: clock.elapsed().as_micros() as u32,
+            timestamp: self.1.elapsed().as_micros() as u32,
             payload_len: payload.len() as u16,
```
AndrewWestberg, line 301: We should check for integer overflow here. It's a protocol error if someone tries to write a segment larger than `65535` bytes.

AndrewWestberg, line 301: We should check for integer overflow here. It's a protocol error if someone tries to write a segment larger than `65535` bytes.

pallas-network/src/multiplexer.rs
```diff
@@ -538,12 +570,12 @@ mod tests {
         minicbor::encode(in_part1, &mut input).unwrap();
         minicbor::encode(in_part2, &mut input).unwrap();
 
-        let ingress = tokio::sync::mpsc::channel(100);
-        let egress = tokio::sync::broadcast::channel(100);
+        let (to_plexer, _) = tokio::sync::mpsc::channel(100);
```
AndrewWestberg, line 573: `const` for channel sizes.

pallas-network/src/multiplexer.rs
```diff
@@ -538,12 +570,12 @@ mod tests {
...
+        let (into_plexer, from_plexer) = tokio::sync::broadcast::channel(100);
 
-        let channel = AgentChannel::for_client(0, &ingress, &egress);
+        let channel = AgentChannel::for_client(0, to_plexer, from_plexer);
 
-        egress.0.send((0x8000, input)).unwrap();
+        into_plexer.send((0x8000, input)).unwrap();
```
AndrewWestberg, line 578: Make `0x8000` a `const` since it's used several places.

AndrewWestberg (approved):
Good updates. Just a few small comments.

pallas-network/src/multiplexer.rs
scarmuega: the `as 32` makes sure that we never overflow. If it goes higher than the u32::MAX, the value will hold the least significant bytes (it would look like a restart)

PR #370, opened 2024-01-02
fix: update pallas-applying to work with keepraw native scripts

jmhrpr:
I think I might have missed this when I made native scripts KeepRaw, feel free to change this if there is a better way to make it work


MaicoLeberle:
I think this solution should do just fine! In any case, I'm working on the Alonzo era at the moment and, for the sake of homogeneity in the code base, I might make other changes on this same file. When I'm ready to merge the current development for the Alonzo era into main, I'll take care of any merging conflicts, should there be any, so don't worry!

scarmuega (approved):

PR #378, opened 2024-01-12
fix(codec) Fix flat encoding and decoding of arbitrarily size integers

KtorZ:
This commits fixes the flat encoding and decoding (and consequently, the zigzag) for large integers in the following ways:

  - It removes support for encoding and decoding i128 values.

  - It optionally (feature = "num-bigint") introduces encoding and decoding of large sized integers through the num-bigint::BigInt type.

  Without the feature enabled, it is still possible to encode and decode isize values; but the use of i128 is now prohibited (as it would overflow on boundaries) in favor of arbitrarily sized integers.

  The commit also introduces a property roundtrip for encoding and decoding large integers, which was missing and thus, failed to identify the overflow problem.

  See related issue: https://github.com/aiken-lang/aiken/issues/796

### DIFF
--- a/pallas-codec/Cargo.toml
+++ b/pallas-codec/Cargo.toml
@@ -14,9 +14,13 @@ authors = [
   "Kasey White <kwhitemsg@gmail.com>",
 ]
 
+[features]
+default = []
+
 [dependencies]
 hex = "0.4.3"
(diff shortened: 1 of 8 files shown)


MicroProofs (approved):
LGTM. Preserves the semantics

PR #411, opened 2024-02-27
Make RollStream fallible and more robust

jmhrpr:


pallas-rolldb/src/wal/stream.rs
```diff
@@ -1,31 +1,49 @@
...
+
+            for entry in iter {
+                let (wal_seq, log) = entry?;
 
-            let iter = store.crawl_from_intersect(&intersect).unwrap();
+                if let Some(prev_seq) = last_seq {
+                    if wal_seq != (prev_seq + 1) {
+                        Err(Error::UnexpectedWalSeq(prev_seq + 1, wal_seq))?
+                    }
+                };
```
scarmuega, line 26: why do we need this check?

jmhrpr: This is what protects against the race condition where the found seq for the intersect has been pruned from the WAL

jmhrpr: Sure, for that exact scenario we would only need to check the first entry returned by the iterator, but I figured why not check all of them

PR #418, opened 2024-03-05
fix: allow optional CBOR set tag in conway

### LINKED ISSUE #410: `InvalidCbor` error when reading blocks from `sanchonet` on `Conway` era
## Context & versions
It looks like the blocks produced by the `8.8.0-pre` Cardano node on the `sanchonet` network for the `Conway` era are not correctly parsed by Pallas `0.23.0`

## Steps to reproduce
We have tried to parse the blocks from latest immutable files produced on the `sanchonet` network with `pallas-hardano` and `pallas-traverse`:

```rust
let chunk = pallas_hardano::storage::immutable::chunk::read_blocks(Path::new("./immutable"), "05110").unwrap();

for block in chunk {
    let block = block.unwrap();
    pallas_traverse::MultiEraBlock::decode(&block).unwrap();
}
```

The immutable files are available on this [Mithril snapshot](https://storage.googleapis.com/cdn.aggregator.testing-sanchonet.api.mithril.network/sanchonet-e257-i5149.7d6ac3f583b1208a6999ddc5626c39087addb1f012a48d573c039c55ec621e2a.tar.zst)

## Actual behavior
We have received the error:
```
called `Result::unwrap()` on an `Err` value: InvalidCbor("unexpected type tag at position 864: expected array")
```

## Expected behavior
The code should not have panicked and parsed a valid `MultiEraBlock`.

jmhrpr:
closes #410


pallas-traverse/src/tx.rs
```diff
@@ -202,13 +202,16 @@ impl<'b> MultiEraTx<'b> {
...
+            MultiEraTx::Conway(x) => {
+                if let Some(ref_inputs) = &x.transaction_body.reference_inputs {
+                    ref_inputs
+                        .iter()
+                        .map(MultiEraInput::from_alonzo_compatible)
+                        .collect()
+                } else {
+                    vec![]
+                }
+            }
```
scarmuega, line 214: why this change?

jmhrpr: Reference inputs was changed to a Set which didn't seem to work with the previous implementation

scarmuega:
closing in favor of #421

PR #419, opened 2024-03-06
feat: optionally overlap WAL with immutable part of chain when pruning

jmhrpr:
https://github.com/txpipe/dolos/issues/173


scarmuega:
@jmhrpr looks good! could we add a unit test to cover the overlap logic?

something like:
- create dummy data
- run purge
- make sure that overlap exists and has correct size

scarmuega (approved):

PR #420, opened 2024-03-08
fix: allow extra bytes when decoding base address

jmhrpr:


jmhrpr:
I thought it might only have affected base addresses, but yes looks like here there are other examples:

https://github.com/IntersectMBO/cardano-ledger/issues/2729

So, we could not allow extra bytes but whitelist the addresses which have been seen on chain, or allow extra bytes for the other address types when we are decoding. What would you prefer?

scarmuega:
what if we add a new type of address `InvalidSpec(bytes)` (or a better name) to the enum?
We could use it as fallback when we find something that is not following the spec. We would keep the raw bytes and avoid making any other assumption regarding the content.

jmhrpr:
The addresses still decoded to a cardano address though, even though there is extra bytes. So i'm not sure it makes sense to not decode them into one of the current options

scarmuega (approved):

scarmuega:
merging as temporal solution. We should make this edge case more explicit in the code somehow.

PR #477, opened 2024-07-01
ci: fix gmp build on windows

scarmuega:
The rug -> gmp dep fails to build on windows because it requires msys2 instead of mvsc.

This PR adjusts the target used in the rust toolchain setup to `stable-x86_64-pc-windows-gnu` which takes care of the setup for msys2, but it still doesn't work because there are package dependencies that need to be installed inside msys2 (which I'm not clear how to do).


scarmuega:
cc @AndrewWestberg

PR #483, opened 2024-07-10
feat: Add support for some math functions

AndrewWestberg:
Enables checks to run under Windows.

Performs the golden tests for the `exp()` function. Addtl math will be added in a later PR. 

The `gmp` feature defines the `Decimal` struct that uses the underlying `gmp` math library.
The `num` feature defines the `Decimal` struct that uses the underlying num-bigint math library.

`gmp` is set as the default and `num` must be enabled when running under Windows. The primary trait for `Decimal` is called `FixedPrecision`. In the future, we could have additional math libraries implement this if they are better/faster, etc...

I'm unsure about making `gmp` a default feature. I'm wondering if it makes more sense to instead create a `pallas-math-gmp` and `pallas-math-num` crate that depend on `pallas-math`. Suggestions are welcome for the best way to structure this.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -1,7 +1,7 @@
 # Based on https://github.com/actions-rs/meta/blob/master/recipes/quickstart.md
 
 on:
-  push: {}
+  push: { }
 
 name: Validate
 
@@ -11,8 +11,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [windows-latest, ubuntu-latest, macOS-latest]
-        rust: [stable]
+        os: [ windows-latest, ubuntu-latest, macOS-latest ]
+        rust: [ stable ]
 
     runs-on: ${{ matrix.os }}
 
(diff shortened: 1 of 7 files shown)


scarmuega:
@AndrewWestberg I agree with the feature-flag strategy. I think that having everything in the same crate but flagged at the root of the lib is a good balance.

What are your thoughts on `num` after this? Is it a reasonable replacement?

.github/workflows/validate.yml
```diff
@@ -59,21 +71,15 @@ jobs:
...
+        run: cargo fmt --all -- --check
 
       - name: Run cargo clippy
-        uses: actions-rs/cargo@v1
-        with:
-          command: clippy
-          args: -- -D warnings
+        run: |
+          cargo clippy -- -D warnings
+          cargo clippy --no-default-features --features num -- -D warnings
```
scarmuega, line 85: @AndrewWestberg was this just for testing? because on main we should cover all features.

AndrewWestberg: This runs clippy twice. Once with the default features (gmp), and once with the `num` feature. The code is designed so that you must enable at least one of the features, but cannot enable both at the same time.

AndrewWestberg: see `lib.rs`

scarmuega (approved):

PR #492, opened 2024-07-23
Include native_bytes in PlutusData

Quantumplation:
Depends on the spec updates [here](https://github.com/utxorpc/spec/pull/111); Technically this doesn't round trip / honor the bytes from the block itself, so could have encoding / hash issues, so we might want to do a larger refactor


scarmuega:
closed in favor of #493

PR #495, opened 2024-07-29
Decode local-TX-submission errors

kettlebell:
This PR replaces `RejectReason(Vec<u8>)` with rust-equivalent error types from the Haskell `cardano-ledger` codebase. In addition I've implemented a CBOR decoder for error responses, including errors that don't fit in a single segment. It can also pass over CBOR bytes of unhandled node errors.

Some errors that can be currently decoded:
- Bad/missing inputs
- Script error
- ValueNotConserved


scarmuega:
Thanks @kettlebell for the PR. We gathered feedback from several PRs and ended up with an implementation that should satisfy requirements for all.

Closing in favor of of #609.

PR #499, opened 2024-08-07
Use Conway types in places they are meant to.

KtorZ:
This properly supersedes #357.

### DIFF
--- a/pallas-primitives/src/conway/model.rs
+++ b/pallas-primitives/src/conway/model.rs
@@ -45,7 +45,7 @@ pub use crate::alonzo::AssetName;
 
 pub type Multiasset<A> = NonEmptyKeyValuePairs<PolicyId, NonEmptyKeyValuePairs<AssetName, A>>;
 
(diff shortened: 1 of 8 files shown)


pallas-primitives/src/conway/model.rs
```diff
@@ -1220,7 +1220,23 @@ where
...
+    #[n(1)]
+    pub value: Value,
+
+    #[n(2)]
+    pub datum_option: Option<T1>,
+
+    #[n(3)]
+    pub script_ref: Option<CborWrap<T2>>,
+}
+
```
scarmuega, line 1238: what's the difference with the babbage one?

KtorZ: Ah no! The `Value` it is!

KtorZ: We could also just parameterize over the Value.

scarmuega (approved):

PR #500, opened 2024-08-07
Add NonceGenerator trait and implement RollingNonce

AndrewWestberg:
The rolling nonce on cardano is calculated for each block. It starts at the shelley genesis hash and applies each block's eta_vrf_0 value to the previous nonce and re-hashing.

Eventually, some of these rolling nonce values are used in the calculation of the epoch nonce which is required for consensus leadership checks.


vincenthz:
This increase the build complexity of pallas substantially and make the build system very brittle, and expose C functions almost as is instead of providing an implementation neutral API.

why not using https://github.com/input-output-hk/vrf ? and probably put all the libsodium stuff behind a feature flag with a capability to switch ?

pallas-crypto/build.rs
```diff
@@ -0,0 +1,92 @@
...
+    // if windows
+    #[cfg(target_os = "windows")]
+    {
+        // Build libsodium with MSBuild
+        run("msbuild", |command| {
+            command
+                .current_dir("contrib/libsodium/builds/msvc/vs2019")
+                .arg("libsodium.sln")
+                .arg("/p:Configuration=StaticRelease")
+                .arg("/p:Platform=x64")
```
vincenthz, line 30: does that mean it will fail when building on arm windows machines ?

NicolasDP: and it looks like we are also losing WASM support here too ?

AndrewWestberg: As far as I can gather, the vrf crate from IOG might not be compatible with libsodium. It seems to be waiting on a PR going into the curve25519_dalek crate that has been open for 2 years without being merged. https://github.com/dalek-cryptography/curve25519-dalek/pull/377

Maybe it is compatible though since it's pointing to the fork of that library. Not sure if that will mess up our dependency graph or not.  Anyway... off to do some testing.

vincenthz: It's meant to be compatible with cardano just like the fork of libsodium that cardano use. it's forking the curve25519 crate and I don't think it should impact the dependencies (or have a build impact as big as depending on a C library).

Don't think it's particularly need any maintenance, which explains the 3years with no update.

AndrewWestberg:
@vincenthz @NicolasDP Please take another look at this PR.

AndrewWestberg:
> This increase the build complexity of pallas substantially and make the build system very brittle, and expose C functions almost as is instead of providing an implementation neutral API.
> 
> why not using https://github.com/input-output-hk/vrf ? and probably put all the libsodium stuff behind a feature flag with a capability to switch ?

This has been implemented. libsodium dependency is gone.

pallas-crypto/src/nonce/mod.rs
```diff
@@ -0,0 +1,17 @@
...
+#[derive(Error, Debug)]
+pub enum Error {
+    #[error("Nonce error: {0}")]
+    Nonce(String),
+}
+
+/// A trait for generating nonces.
+pub trait NonceGenerator: Sized {
+    fn finalize(&mut self) -> Result<Hash<32>, Error>;
+}
```
NicolasDP, line 17: Can we change a bit the trait and remove this Error type that is not carrying any useful meaning in this case ?

How about going to a more generic approach:

```rust
pub trait NonceGenerator {
    type Output;

     fn finalize(&mut self) -> Self::Output;
 }
```

That way we can remove the error handling that is not necessary for the EpochGenerator:

```rust
impl NonceGenerator for EpochNonceGenerator {
    type Output = Hash<32>;

     fn finalize(&mut self) -> Self::Output {
         Ok(self.nonce)
     }
 }
```

And have an error handling more precise for the `RollingNonceGenerator`:

```rust
#[derive(Error)]
#[error("Must call apply_block before finalise!")]
struct RollingNonceGeneratorError;

impl NonceGenerator for RollingNonceGenerator {
    type Output = Result<Hash<32>, RollingNonceGeneratorError>;

     fn finalize(&mut self) -> Self::Output {
        todo!()
     }
 }
```

What do you think? Would that make sense on the overall picture ? I like it in the sense that (though I recognise I am bias for this kind of API) it gives more explicit details of what the Error handling is (and no error handling if not needed) and also allows the opportunity to remove the need to stick to `Hash<32>` (allow to use `Hash<24>` for example).

scarmuega: Do we need a trait at all?

pallas-crypto/Cargo.toml
```diff
@@ -18,7 +21,12 @@ rand_core = "0.6"
 pallas-codec = { version = "=0.30.2", path = "../pallas-codec" }
 serde = "1.0.143"
 
+# The vrf crate has not been fully tested in production environments and still has several upstream issues that
+# are open PRs but not merged yet.
+vrf_dalek = { git = "https://github.com/input-output-hk/vrf", rev = "a3185620b72e6a9647285941b961021186f16693" }
```
scarmuega, line 26: @NicolasDP @AndrewWestberg PR seems to have been merged now.

Regardless, I'm worried that there's no license applied in the upstream repo. Is that something that could be added? Should I open an issue?

pallas-crypto/src/vrf/mod.rs
```diff
@@ -0,0 +1,74 @@
...
+    #[error("TryFromSlice {0}")]
+    TryFromSlice(#[from] std::array::TryFromSliceError),
+
+    #[error("VrfError {0}")]
+    VrfError(#[from] vrf_dalek::errors::VrfError),
+}
+
+/// Sign a seed value with a vrf secret key and produce a proof signature
+pub fn vrf_prove(secret_key: &[u8], seed: &[u8]) -> Result<Vec<u8>, Error> {
+    let sk = SecretKey03::from_bytes(secret_key[..32].try_into()?);
```
NicolasDP, line 15: disclaimer: _I have nothing against free functions_ (™️ )

why not adding new types for the `SecretKey` and `VerificationKey` ? That way you can wrap up the vrf_dalek dependency and eventually allow us to switch things around without touching the APIs too much.

also would allow us to add some more conversion API (like the `try_into` but also a memory scrub on drop).

pallas-crypto/src/nonce/mod.rs
NicolasDP: Indeed, I don't think this is needed.

AndrewWestberg: I'll remove and keep it simple with just some functions.

pallas-crypto/Cargo.toml
AndrewWestberg: When I spoke to Sam Leathers, he talked about moving this repository to intersect. I would open it up as an issue and tag him. There's also the KES repository we'd want to do the same thing with.

pallas-crypto/src/nonce/mod.rs
AndrewWestberg: This is ready to re-check.

pallas-crypto/src/vrf/mod.rs
AndrewWestberg: I've added the wrappers VrfSecretKey and VrfPublicKey. Also added the scrub on drop for the SecretKey and some From traits.

pallas-crypto/src/vrf/mod.rs
```diff
@@ -0,0 +1,159 @@
+use crate::hash::Hash;
+use crate::memsec::Scrubbed;
+use thiserror::Error;
+use vrf_dalek::vrf03::{PublicKey03, SecretKey03, VrfProof03};
+
+#[derive(Error, Debug)]
+pub enum Error {
+    #[error("VrfError {0}")]
+    VrfError(#[from] vrf_dalek::errors::VrfError),
+}
```
NicolasDP, line 10: This is me being picky again, the error type could be simplified and we should take ownership of the error handling:

This error is only for the verification (and I like that we have one error for the VRF Verification and not one for both the PublicKey decoding, PrivateKey decoding etc...).

How about we make the error a bit more explicit to clarifify the API:

```suggestion
/// error that can be returned if the verification of a [`VrfProof`] fails
/// see [`VrfProof::verify`]
///
#[derive(Error, Debug)]
#[error("VRF Proof Verification failed.")]
pub struct VerificationError(
    #[from] // so we keep the `?` handling in the function as you used
    #[source] // it is included with the `#[from]` so you can ignore tbh
    vrf_dalek::errors::VrfError
);
```

pallas-crypto/src/vrf/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+        VrfPublicKey {
+            public_key_03: PublicKey03::from(&secret_key.secret_key_03),
+        }
+    }
+}
+
+// Wipe the secret key
+impl Scrubbed for VrfSecretKey {
+    fn scrub(&mut self) {
+        self.secret_key_03.to_bytes().scrub();
```
NicolasDP, line 72: we aren't really wiping the secret key here:

https://github.com/input-output-hk/vrf/blob/a3185620b72e6a9647285941b961021186f16693/src/vrf03.rs#L32-L35

The `to_bytes` functions returns a copy of the byte array of the secret key. It's still in memory. I wonder if we can either change that upstream (better choice) or be cheeky (not great) here:

```suggestion
        let mut bytes: [u8; VRF_SECRET_KEY_SIZE] = std::men::transmute(&mut self.secret_key_03);
```

I think the upstream change would make more sense but that's also a direction that can be taken.

pallas-math/src/math_gmp.rs
```diff
@@ -279,6 +317,18 @@ impl FixedPrecision for Decimal {
...
+        todo!()
+    }
+
+    fn floor(&self) -> Self {
+        todo!()
+    }
+
+    fn ceil(&self) -> Self {
+        todo!()
+    }
```
NicolasDP, line 331: are you leaving the TODOs for another PR?

AndrewWestberg: math_gmp.rs is going to be removed before I merge. Too many headaches with it.

pallas-crypto/src/vrf/mod.rs
AndrewWestberg: https://github.com/input-output-hk/vrf/pull/3

pallas-math/src/math_gmp.rs
AndrewWestberg: It's gone.

pallas-crypto/src/vrf/mod.rs
AndrewWestberg: upstream fix is merged.

pallas-crypto/src/nonce/mod.rs
```diff
@@ -0,0 +1,192 @@
...
+        hasher.input(extra_entropy);
+        hasher.finalize()
+    } else {
+        epoch_nonce
+    }
+}
+
+/// A nonce generator function that calculates a rolling nonce (eta_v) by applying each cardano block in
+/// the shelley era and beyond. These rolling nonce values are used to help calculate the epoch
+/// nonce values used in consensus for the Ouroboros protocols (tpraos, praos, cpraos).
```
NicolasDP, line 22: This function may panic and I think it would make sense to document this at the function's documentation level.

```suggestion
/// nonce values used in consensus for the Ouroboros protocols (tpraos, praos, cpraos).
///
/// # Panic
///
/// This function may panic if the `block_eta_vrf_0` argument is not a slice of
/// either 32 bytes or 64 bytes.
///
```

pallas-math/src/math.rs
```diff
@@ -49,6 +46,18 @@ pub trait FixedPrecision:
...
+    fn round(&self) -> Self;
+
+    /// Round down to the nearest integer number
+    fn floor(&self) -> Self;
+
+    /// Round up to the nearest integer number
+    fn ceil(&self) -> Self;
+
+    /// Truncate to the nearest integer number
+    fn trunc(&self) -> Self;
```
NicolasDP, line 60: Usually API like this also comes with a `must_use` warning. This helps the compiler warns the user if they are expecting to use the output of the function as the function does not modify the `self` in place.

For example: https://doc.rust-lang.org/1.81.0/src/core/num/mod.rs.html#546

```suggestion
    /// Round to the nearest integer number
    #[must_use]
    fn round(&self) -> Self;

    /// Round down to the nearest integer number
    #[must_use]
    fn floor(&self) -> Self;

    /// Round up to the nearest integer number
    #[must_use]
    fn ceil(&self) -> Self;

    /// Truncate to the nearest integer number
    #[must_use]
    fn trunc(&self) -> Self;
```

NicolasDP (approved):
Just last two comments (that should be able to be added via the GitHub API if you wish to add them in). Otherwise I give my stamp of approval 🥇 .

Thanks so much for this great contribution to the pallas -- and to the Cardano Rust -- ecosystem.

NicolasDP (approved):

PR #512, opened 2024-09-08
Add script_data_hash computation.

kalomaaan:


pallas-txbuilder/src/transaction/model.rs
```diff
@@ -340,8 +340,41 @@ impl StagingTransaction {
...
+
+        let mut buf = Vec::new();
+        if redeemers.len() == 0 && datums.is_some() {
+            buf.push(0xA0);
```
AndrewWestberg, line 350: This is the value that changed from `0x80` -> `0xA0` for Conway Era, correct?

pallas-txbuilder/src/transaction/model.rs
```diff
@@ -340,8 +340,41 @@ impl StagingTransaction {
...
...00, 53, 25, 3, 1, 4, 2, 26, 0, 1, 79, 88, 26, 0, 3, 124, 113, 24, 122, 0, 1, 1, 25, 3, 232, 25, 167, 169, 4, 2, 25, 95, 228, 25, 115, 58, 24, 38, 1, 26, 0, 13, 180, 100, 25, 106, 143, 1, 25, 202, 63, 25, 2, 46, 1, 25, 153, 16, 25, 3, 232, 25, 236, 178, 1, 26, 0, 2, 42, 71, 24, 32, 26, 0, 1, 68, 206, 24, 32, 25, 59, 195, 24, 32, 26, 0, 1, 41, 17, 1, 25, 51, 113, 4, 25, 86, 84, 10, 25, 113, 71, 24, 74, 1, 25, 113, 71, 24, 74, 1, 25, 169, 21, 25, 2, 40, 1, 25, 174, 205, 25, 2, 29, 1, 25, 132, 60, 24, 32, 26, 0, 1, 10, 150, 24, 32, 26, 0, 1, 26, 170, 24, 32, 25, 28, 75, 24, 32, 25, 28, 223, 24, 32, 25, 45, 26, 24, 32, 26, 0, 1, 79, 88, 26, 0, 3, 124, 113, 24, 122, 0, 1, 1, 26, 0, 1, 97, 66, 25, 2, 7, 0, 1, 26, 0, 1, 34, 193, 24, 32, 26, 0, 1, 79, 88, 26, 0, 3, 124, 113, 24, 122, 0, 1, 1, 26, 0, 1, 79, 88, 26, 0, 3, 124, 113, 24, 122, 0, 1, 1, 26, 0, 4, 33, 60, 25, 88, 60, 4, 26, 0, 22, 60, 173, 25, 252, 54, 4, 25, 79, 243, 1, 4, 0, 26, 0, 2, 42, 168, 24, 32, 26, 0, 1, 137, 180, 25, 1, 164, 1, 1, 26, 0, 1, 62, 255, 24, 32, 25, 232, 106, 24, 32, 25, 78, 174, 24, 32, 25, 96, 12, 24, 32, 25, 81, 8, 24, 32, 25, 101, 77, 24, 32, 25, 96, 47, 24, 32, 26, 3, 46, 147, 175, 25, 55, 253, 10, 255];
```
AndrewWestberg, line 344: Why are these values hardcoded? They should come from the costmodels in protocol params.

pallas-txbuilder/src/transaction/model.rs
Quantumplation: yes, and is just a side effect of encoding an empty map of redeemers.

To correctly handle this, the different era implementations of `redeemers` should have era-specific serialization code; because Chang+1 will mean the redeemers now must be encoded as a map.

pallas-txbuilder/src/transaction/model.rs
kalomaaan: i cant find anything inside pallas on how to get cost models, this should be constant right?, also cost models ends up in language view, i've had @hrpr help me on how to get the correct language view so i figured to hard code it to avoid the hassle.

AndrewWestberg: cost models are an updatable protocol parameter that can change at any time. They should be retrievable from a live node in n2c mode doing a protocol parameters query.

image

Quantumplation: In particular, for this patch, they should be passed in to the script data hash function. I *think* Pallas already has types for the cost models over where the protocol parameters are defined?

pallas-txbuilder/src/transaction/model.rs
scarmuega: instead of handling cbor bytes directly, it would be easier / safer to have a struct `ScriptDataHash` that implements cbor encoding and contains the underlying data parts (empty maps and such nuances would be handled by the data type).

This struct could live inside primitives crate and get versioned by era (`babbage:ScriptDataHash`, `conway:ScriptDataHash`, etc).

scarmuega:
closing in favor of #525

PR #517, opened 2024-10-03
feat(crypto): Add extra types and conversions

AndrewWestberg:
Adding some ergonomic improvements to make Amaru code less ugly.

### DIFF
--- a/pallas-codec/src/utils.rs
+++ b/pallas-codec/src/utils.rs
@@ -4,6 +4,7 @@ use minicbor::{
     Decode, Encode,
 };
 use serde::{Deserialize, Serialize};
+use std::str::FromStr;
 use std::{collections::HashMap, fmt, hash::Hash as StdHash, ops::Deref};
 
 static TAG_SET: u64 = 258;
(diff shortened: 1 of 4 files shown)


pallas-primitives/src/babbage/model.rs
```diff
@@ -313,6 +313,34 @@ impl<'a> From<MintedTransactionBody<'a>> for TransactionBody {
...
+impl HeaderBody {
+    pub fn leader_vrf_output(&self) -> Vec<u8> {
+        derive_tagged_vrf_output(&self.vrf_result.0, VrfDerivation::Leader)
+    }
+
+    pub fn nonce_vrf_output(&self) -> Vec<u8> {
+        derive_tagged_vrf_output(&self.vrf_result.0, VrfDerivation::Nonce)
+    }
+}
+
```
scarmuega, line 343: I'm not sure if we should put this logic in primitives. Why are we moving it from traverse to here?

pallas-codec/src/utils.rs
```diff
@@ -1340,6 +1341,12 @@ impl From<Bytes> for Vec<u8> {
     }
 }
 
+impl<const N: usize> From<&Bytes> for [u8; N] {
+    fn from(b: &Bytes) -> Self {
+        b.0.as_slice()[..N].try_into().expect("Infallible")
+    }
+}
```
scarmuega, line 1348: I don't think we can consider this conversion infallible. Wouldn't this panic if bytes len is less that N?

AndrewWestberg: Yeah. I'll convert it to TryFrom.

pallas-primitives/src/babbage/model.rs
AndrewWestberg: I moved it here because we said we didn't want to use pallas-traverse in Amaru.  I believe it belongs in primitives because a leader_vrf and nonce_vrf exist in Alonzo and earlier headers. In babbage and beyond, these values are calculated instead of included in the header itself. I would consider these values as part of the header even though they're not explicitly in the header.

pallas-codec/src/utils.rs
AndrewWestberg: fixed

pallas-primitives/src/babbage/model.rs
scarmuega: I approved and merged the PR to move along, but I think that adding business logic to a specific era in the primitives crate is a slippery slope. I admit that this particular case is pretty constrained and harmless, but I just want to state that the original scope of the primitives crate is to hold data structures matching the CDDL and their corresponding encoding / decoding logic.

@KtorZ advocates for leaving `pallas-traverse` outside of the shared kernel, I'm still on the fence. What I don't like is slowly merging business logic into primitives just to accommodate for this decision.

Pallas implementors need an api layer that is era-agnostic and more ergonomic than CDDL structures. If Amaru doesn't want to use pallas-traverse, then I believe this api layer should be Amaru-specific and live within it's own repo.

PR #518, opened 2024-10-09
feat(crypto): Add Key Evolving Signatures (KES)

AndrewWestberg:
This code modifies block headers so that the HeaderBody bytes are wrapped in KeepRaw. These raw bytes are used in the validation of KES signatures.

pallas-crypto has been updated with wrappers around kes signatures for both evolving the secret key as well as verifying signatures with the public key.

### DIFF
--- a/pallas-crypto/src/kes/mod.rs
+++ b/pallas-crypto/src/kes/mod.rs
@@ -1,0 +1,134 @@
+use kes_summed_ed25519::kes::{Sum6Kes, Sum6KesSig};
+use kes_summed_ed25519::traits::{KesSig, KesSk};
+use kes_summed_ed25519::PublicKey;
+use thiserror::Error;
+
+/// KES error
+#[derive(Error, Debug)]
+pub enum Error {
+    #[error("KES error: {0}")]
+    Kes(#[from] kes_summed_ed25519::errors::Error),
+}
+
+/// KES secret key
+pub struct KesSecretKey<'a> {
+    sum_6_kes: Sum6Kes<'a>,
+}
+
+impl KesSecretKey<'_> {
+    /// Create a new KES secret key
+    pub fn from_bytes(sk_bytes: &mut Vec<u8>) -> Result<KesSecretKey, Error> {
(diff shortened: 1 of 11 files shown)


pallas-crypto/src/kes/mod.rs
```diff
@@ -0,0 +1,128 @@
...
+
+/// KES error
+#[derive(Error, Debug)]
+pub enum Error {
+    #[error("KES error: {0}")]
+    Kes(#[from] kes_summed_ed25519::errors::Error),
+}
+
+/// KES secret key
+pub struct KesSecretKey<'a> {
```
scarmuega, line 14: Is there any scrubbing mechanism in the upstream repo that we can use on Drop to ensure we clear private key data from memory?

pallas-crypto/src/kes/mod.rs
```diff
@@ -0,0 +1,128 @@
...
+        // TODO: extend() could potentially re-allocate memory to a new location and copy the sk_bytes.
+        // This would leave the original memory containing the secret key without being wiped.
+        sk_bytes.extend([0u8; 4]); // default to period = 0
+        let sum_6_kes = Sum6Kes::from_bytes(sk_bytes.as_mut_slice())?;
+        Ok(KesSecretKey { sum_6_kes })
+    }
+
+    /// Get the internal representation of the KES secret key at the current period
+    /// This value will include the period as the last 4 bytes in big-endian format
+    pub fn as_bytes(&self) -> &[u8] {
```
scarmuega, line 30: In other parts of the crypto crate we use the name `leak_into_bytes` and add a Safety disclaimer to make sure devs understand the risks of exposing these bytes. Maybe we could do the same here.

pallas-crypto/src/kes/mod.rs
AndrewWestberg: Yes. I added this here: https://github.com/txpipe/kes/pull/2

pallas-crypto/src/kes/mod.rs
AndrewWestberg: fixed.

scarmuega (approved):

PR #520, opened 2024-10-14
refactor: improved support for Conway genesis and protocol params

Mercurial:


scarmuega:
great work!
I just prefer to keep the macros simpler and treat cost-models as special cases.
Closing in favor of #521

PR #523, opened 2024-10-18
Re-organize and clean-up pallas-primitives

KtorZ:
Namely:

  - Move _common_ (i.e. era-independent) types and structures up to the `lib` module; to be shared across all eras. If any of those deviate in a subsequent era, it is easy to bring them down and define new types from the point of divergence onward. This simplifies the scope of each era-specific module and make them slightly easier to navigate.

    Note that, each era module still re-export all of the common types that's relevant to that particular era. So technically, this reorganization doesn't really change anything for callers/users of the library.

  - Rename `Scripthash` to `ScriptHash`. Before this commit, both actually existed as `ScriptHash` was introduced with the Conway era. Yet, they refer to the same thing, so the duplication is simply confusing.

  - Rename `One` / `Two` constructors for `NetworkId` to `Testnet` and `Mainnet` respectively. Also defined idiomatic `From` & `TryFrom` implementation for conversion to and from `u8`. This is a lot let confusing!

  - Generalize `PlutusScript` with a constant generic, to avoid repetition for each plutus script generated for specific version. Note that a distinction is still _necessary_ if we want to provie out-of-the-box serialisers for Plutus scripts, which are serialised with a tag prefix depending on the language. All else apart, they are strictly similar types.

  - Rename `CostMdls` to `CostModels`. Because, common.

  - Rename `plutus_script` to `plutus_v1_script` in the Alonzo's witness set, for consistency with other eras.


pallas-primitives/src/lib.rs
```diff
@@ -1,10 +1,392 @@
...
+    }
+}
+
+pub type MetadatumLabel = u64;
+
+#[derive(
+    Serialize, Deserialize, Encode, Decode, Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Copy,
+)]
+#[cbor(index_only)]
+pub enum NetworkId {
```
AndrewWestberg, line 139: This is highly cardano-specific. APEX Fusion networks use the following

```
prime testnet: networkId == 0
prime mainnet: networkId == 1
vector testnet: networkId == 2
vector mainnet: networkId == 3
```

Does this code limit us to using only these specific values with pallas?

AndrewWestberg (approved):

pallas-primitives/src/lib.rs
KtorZ: Not sure to see the relevance of APEX here? I mean, 90% of the content of pallas-primitives is Cardano-specific 😅 ...

pallas-primitives/src/alonzo/model.rs
```diff
@@ -1308,7 +702,11 @@ pub struct WitnessSet {
     pub bootstrap_witness: Option<Vec<BootstrapWitness>>,
 
     #[n(3)]
-    pub plutus_script: Option<Vec<PlutusScript>>,
+    pub plutus_v1_script: Option<Vec<PlutusScript<1>>>,
+
+    #[cbor(skip)]
+    #[deprecated(since = "0.31.0", note = "use `plutus_v1_script` instead")]
+    pub plutus_script: Option<Vec<PlutusScript<1>>>,
```
scarmuega, line 709: Do we need this change?

This is Alonzo specific and the CDDL names this field as `plutus_script`.

I'm worried that having a field skipped by cbor decoding will be a hard issue to debug.

pallas-primitives/src/lib.rs
scarmuega: If we kept this as Alonzo-specific, I would follow Alonzo CDDL and constrains this to 0 / 1.

Since we're moving this a more generic primitive, I would turn this into an int and leave the semantics of each value to a higher layer.

pallas-primitives/src/plutus_data.rs
```diff
@@ -0,0 +1,353 @@
...
+        e.array(0)?;
+    } else {
+        e.begin_array()?;
+        for v in a {
+            e.encode_with(v, ctx)?;
+        }
+        e.end()?;
+    }
+    Ok(())
+}
```
scarmuega, line 353: I don't mind this, but maybe better suited for pallas::codec::utils as a newtype (`HardanoList` or better name)?

pallas-primitives/src/alonzo/model.rs
KtorZ: Names don't exist in CBOR; it's only annotation in CDDL so we could pick whatever. I chose to rename it to `plutus_v1_script` for consistency with other eras. I can revert, but as the codebase evolves, I rather keep things tidy than carrying over oddities from the past as it only creates more cognitive overload. 

The skipped field here is only added per your request to not break the API and provide clear deprecation warnings. It is perhaps overkill for struct field? Happy to just drop it.

pallas-primitives/src/plutus_data.rs
KtorZ: I have intention to rewrite a bit some of the data encoder here; because the underlying type for `Constr` in particular has proven quite unpractical to use. I understand why the implementation ended that way, but I believe we can do better! 

So for now, I've only parked the Plutus Data-related code in a separate module but haven't tweaked anything. So, this code has just been moved around. This encoding in particular is specific to Plutus data and seen nowhere else in the Cardano codebase; so I believe it better lives with the rest of the Plutus Data stuff.

pallas-primitives/src/alonzo/model.rs
scarmuega: Thanks for explaining how CBOR and CDDL works.

Regardless, I once again state that the goal of the primitives crate is to reflect the CDDL as much as possible so that it might, at some point, be autogenerated (or at least partially).

The primitives crate is NOT the place for ergonomic improvements. For nice names and ergonomic types, we can use traverse or a custom wrapper in Amaru.

pallas-primitives/src/plutus_data.rs
scarmuega: I thought this was a new fn.
Ok, lets keep it until a further refactor.

pallas-primitives/src/alonzo/model.rs
rvcas: hm I think trying to reflect the CDDL is silly and I think autogenerating is a pipe dream

Quantumplation: Maybe we can clarify the CDDL upstream? I was hoping to submit a PR, because in the latest CDDL they've removed the helpful comments that describe what each field in the transaction body is anyway lol: 

https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/conway/impl/cddl-files/conway.cddl#L451-L470

scarmuega: > hm I think trying to reflect the CDDL is silly and I think autogenerating is a pipe dream. That said, I'm largely indifferent. But I can say from experience, trusting KtorZ with these kinds of choices is always worth it long term.

I respect your opinion @rvcas and I trust @KtorZ expertise / track-record, but I fail to see why we want to introduce a breaking change that would affect 15+ downstream repos for an arbitrary name change that, IMO, is against the goal of the module.

These CDDLs are as close as we can get to a spec in Cardano. Either via codegen or manually, there needs to be a direct, 1:1 mapping between the spec and the decoding logic. Adding an extra layer of arbitrary naming translation in the serialization layer makes the module much harder to maintain.

I've been maintain this codebase for a few years now. I've been woken at night several times now because some [peculiar block](https://github.com/txpipe/pallas/blob/4871342a8deffaceafb5e0f771e1623dfe57e25f/pallas-primitives/src/alonzo/model.rs#L1615) had a CBOR deserialization error. Having the codec primitives as similar to the CDDL made it easier to find the root cause. I'll be more than happy to introduce any breaking change that brings the two closer together.

I'm also in favor of adding an ergonomic API layer on top of the primitives. That's what the `traverse` crate is for, but I won't oppose to someone building a different one if the design decisions are incompatible.

I try to welcome all contributions and I consider everybody in this discussion more knowledgeable than me, but please don't disregard my feedback as silly because it's coming from direct experience maintaining this codebase.

rvcas: No worries, I wasn't disregarding you or your opinion when I said that. Apologies if it came off negatively. If these were voice messages the tone of my voice would have indicated this but apologies anyways for not being more careful in text format.

KtorZ: I believe the right move is what @Quantumplation hinted at:

> Maybe we can clarify the CDDL upstream?

As for the motivation behind this change, simply consistency, which in long run pays off towards maintainability and discoverability. Anyway, since this is sparkling too much debate and we're already at 4000 words spilled for a 3 character change in a now-legacy era, I am going to revert that one!

PR #535, opened 2024-10-31
fix(math): update once_cell::Lazy -> std::sync::LazyLock

AndrewWestberg:
LazyLock is now the preferred way of doing lazy static initialization.


rvcas:
Good to know

scarmuega (approved):

PR #543, opened 2024-11-06
explore: codegen primitives from cddl

scarmuega:


scarmuega:
Closing due to inactivity — this was marked `explore:` and hasn't moved since Nov 2024. Happy to revive as a fresh PR if we want to pick the experiment back up.

PR #548, opened 2024-11-21
fix(network): drop the `[2,…]` message tag wrapper from the rejection reason

michalrus:
## The bug

Currently, the rejection reason is always prefixed with `8202`, because the entire `d.input()` is copied there. And `d.input()` always contains the whole input.

An (incorrect) example of a `FeeTooSmallUTxO` error:

```cbor
82                                      # array(2)
   02                                   # unsigned(2)
   81                                   # array(1)
      82                                # array(2)
         06                             # unsigned(6)
         82                             # array(2)
            82                          # array(2)
               01                       # unsigned(1)
               82                       # array(2)
                  00                    # unsigned(0)
                  83                    # array(3)
                     06                 # unsigned(6)
                     1B 000000025050D4DD # unsigned(9937409245)
                     1A 003D0E39        # unsigned(4001337)
            82                          # array(2)
               01                       # unsigned(1)
               82                       # array(2)
                  00                    # unsigned(0)
                  83                    # array(3)
                     05                 # unsigned(5)
                     1A 0002851D        # unsigned(165149)
                     19 0539            # unsigned(1337)
```

This is not correct, because the first 2 bytes representing `[2, …]` are just a tag for `Message::RejectTx`, which we already give to the user, so the information is currently encoded twice.

The above example should have started on line 3 (3rd byte).

## The fix

We need to:

* offset `d.input()` in `RejectReason` by the current `d.position()`,

* I've also moved `d.set_position(_)` – it's the cleanest to have it in the decoder for `RejectReason`, so that it behaves like other well-formed decoders, i.e. increments the current `.pos`.

### DIFF
--- a/pallas-network/src/miniprotocols/localtxsubmission/codec.rs
+++ b/pallas-network/src/miniprotocols/localtxsubmission/codec.rs
@@ -42,11 +42,6 @@ impl<'b, Tx: Decode<'b, ()>, Reject: Decode<'b, ()>> Decode<'b, ()> for Message<
             // if the first element isn't an array, it's a plutus error
             // the node sends string data
             let rejection = d.decode()?;
-
-            // skip this data via setting the decoder position, because it doesn't recognize
-            // it with rejection decode
-            d.set_position(d.input().len());
-
             return Ok(Message::RejectTx(rejection));
         }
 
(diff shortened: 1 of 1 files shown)


scarmuega (approved):

michalrus:
Thanks for the merge!

Issue #496, opened 2024-08-01
queries_v16::get_current_pparams fails on conway preview network


jonathangenlambda:
Cardano preview has been hard forked into Conway.

When running `queries_v16::get_current_pparams` against a locally synced preview node with version 9.1 we get the following failure 
```InvalidCbor(Error { err: TypeMismatch(Array), pos: Some(52), msg: "expected u64" })```

The era param has value 6 and was successfully fetched via `queries_v16::get_current_era`

PR #551, opened 2024-11-28
fix(network): correct PoolDistr and update ProtocolParam

sterraf:
A field of the `Pool` struct (corresponding to the node's `PoolDistr`) was a naively encoded rational number `Fraction`; this one is indeed needed in the genesis config struct, which I renamed from `Genesis` to the more explicit `GenesisConfig`. 

In the case of `ProtocolParam`, the `ProtocolVersion` is now a tuple `(major, minor)`.

Both corrections have an impact on the `examples/n2c-miniprotocols` folder; two broken examples there now run smoothly. In that folder, I moved the initial `Point` of the block-fetching performed by `do_chainsync` to a more recent moment in order to be able to witness the reaching of the chain tip.

This solves #496.

### DIFF
--- a/pallas-network/src/miniprotocols/localstate/queries_v16/mod.rs
+++ b/pallas-network/src/miniprotocols/localstate/queries_v16/mod.rs
@@ -111,6 +111,7 @@ pub type PositiveInterval = RationalNumber;
 
 pub type ProtocolVersionMajor = u64;
 pub type ProtocolVersionMinor = u64;
+pub type ProtocolVersion = (ProtocolVersionMajor, ProtocolVersionMinor);
 
 pub type CostModel = Vec<i64>;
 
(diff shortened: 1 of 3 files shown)


pallas-network/src/miniprotocols/localstate/queries_v16/mod.rs
```diff
@@ -201,19 +200,21 @@ pub struct StakeDistribution {
...
+/// as an untagged array.
 #[derive(Debug, Encode, Decode, PartialEq, Clone)]
 pub struct Fraction {
     #[n(0)]
     pub num: u64,
 
     #[n(1)]
-    pub dem: u64,
+    pub den: u64,
 }
```
scarmuega, line 218: Do we need this struct now that we're using `RationalNumber` instead?

sterraf:
I believe so, because the encoding of the fraction at `GenesisConfig` is
not the same (!). So it seems two different types are needed.

pallas-network/src/miniprotocols/localstate/queries_v16/mod.rs
sterraf: FYI Reply is below

scarmuega (approved):

PR #553, opened 2024-12-09
making changes to tx builder model

logicalmechanism:
Adding a signature to a conway era transaction causes a `called `Result::unwrap()` on an `Err` value: CorruptedTxBytes` error.

```rust
let signed_tx_cbor = tx
    // called `Result::unwrap()` on an `Err` value: CorruptedTxBytes
    .add_signature(witness_public_key, witness_vector)
    .unwrap()
    .tx_bytes;
```

This error comes from line 685 in model.rs of the txbuilder.

If you change the `babbage` call into a `conway` call then it does not throw the error. 

This PR changes the `babbage` to `conway` on all the things. Also, it adjusts the witness building as pushing to a set is not possible and requires coverting back to a vector then back to an empty set.

This PR passes all current tests. There is no open issue on this topic as it brought up in discord.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,6 @@
 <a name="unreleased"></a>
 ## [Unreleased]
 
-- **txbuilder:** use conway types for decode fragments during `add_signature` ([#xxx](url here))
-
 <a name="v0.30.1"></a>
 ## [v0.30.1] - 2024-08-25
 ### Fix
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 <a name="unreleased"></a>
 ## [Unreleased]
 
+- **txbuilder:** use conway types for decode fragments during `add_signature` ([#xxx](url here))
 
 <a name="v0.30.1"></a>
 ## [v0.30.1] - 2024-08-25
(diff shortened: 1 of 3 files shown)


Mercurial (approved):

logicalmechanism:
I also commented out a debug statement that clutters the terminal screen.

Mercurial (approved):

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 <a name="unreleased"></a>
 ## [Unreleased]
 
+- **txbuilder:** use conway types for decode fragments during `add_signature` ([#xxx](url here))
```
scarmuega, line 4: changelogs get added automatically on release, no need to add them in the PR

pallas-txbuilder/src/conway.rs
```diff
@@ -230,7 +230,8 @@ impl BuildConway for StagingTransaction {
                 language_view,
             };
 
-            dbg!(&dta);
+            // if we need to debug then uncomment
+            // dbg!(&dta);
```
scarmuega, line 234: you can safely remove that, I was meaning to remove it myself (was left there unintentionally)

logicalmechanism: removed

scarmuega (approved):

PR #558, opened 2024-12-13
added derive clone to the required structs for StagingTransaction

logicalmechanism:
Each field of a StagingTransaction and BuiltTransaction should derive Clone. I tested this on-chain, and it is working as expected. All tests are passing.


logicalmechanism:
A duplicate `use hex::FromHex;` is removed too.

scarmuega (approved):

PR #562, opened 2024-12-17
feat(network): Partially structuring RejectReason

sterraf:
This PR decodes the `RejectReason` after a failed local Tx submission by structuring into two variants:

1. Plutus errors (an ASCII string); or
2. an `(Era, Vec<TxError>)` pair.

The `TxError` type, in turn, will be further structured in subsequent PRs. As of now, it has an `ExtraneousScriptWitnessesUTXOW` variant corresponding to [the appropriate `ConwayUtxowPredFailure`](https://github.com/IntersectMBO/cardano-ledger/blob/12f6aa6f094af5dab722edf03d4ab3f4ec99aa48/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Utxow.hs#L102) , an `u8` variant, and finally a raw variant holding cases that have not been analyzed yet.

A unit test using an up-to-date preprod node response was added to `codec.rs`.


sterraf:
I've recently learned about the related #495. It might be necessary to determine its status.

pallas-network/src/miniprotocols/localtxsubmission/codec.rs
```diff
@@ -101,24 +104,77 @@ impl Encode<()> for EraTx {
...
+            _ => {
+                for _ in 1..size {
+                    d.skip()?
+                } // Skip decoding the elements of the array
+                  // Store the slice of raw CBOR
+                raw_elements.append(&mut data[start_pos..d.position()].to_vec());
+
+                Ok(TxError::Raw(raw_elements))
+            }
+        }
```
scarmuega, line 174: this seems a little bit more "imperative" than what I was hoping for. Instead embedding the decoding logic as conditional branches, let the decoder do it's thing by defining intermediate structures that handle smaller chunks of the logic.

sterraf: Thank you for the insight; I added a missing layer of the error structure. Hope it looks better now!

sterraf:
Subsumed by #598

PR #563, opened 2024-12-18
fix: add min_fee_sript_ref_cost_per_byte to parameter mapper

ricomiles:
This adds min_fee_script_ref_cost_per_byte to the parameter mapper, ensuring it is exposed.
It also updates the type of an existing parameter from UnitInterval to Coin to match the type specified in the spec update.

This PR uses a local spec dependency, which needs to be updated once the new spec is released.

### DIFF
--- a/pallas-applying/src/utils/environment.rs
+++ b/pallas-applying/src/utils/environment.rs
@@ -176,7 +176,7 @@ pub struct ConwayProtParams {
     pub governance_action_deposit: Coin,
     pub drep_deposit: Coin,
     pub drep_inactivity_period: Epoch,
-    pub minfee_refscript_cost_per_byte: Coin,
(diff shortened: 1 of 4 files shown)


scarmuega:
the [cddl](https://github.com/IntersectMBO/cardano-ledger/blob/6edd909ca5825bb0b9a335eb637ccffb2b78ab61/eras/conway/impl/cddl-files/conway.cddl#L384) shows the param as a `nonnegative_interval`. Why are we changing it to coin?

ricomiles:
ohh my bad, I overlooked that part, thank you for pointing it out. I'll update it right away

ricomiles:
reverted it back to UnitInterval @scarmuega, 
I will also close this [txpipe/dolos#402](https://github.com/txpipe/dolos/pull/402), since they are related.

PR #574, opened 2025-01-03
feat(network): add `peersharing` protocol module

franciscojoray:
This PR adds the implementation of the `peersharing` protocol.

### DIFF
--- a/pallas-network/src/miniprotocols/peersharing/client.rs
+++ b/pallas-network/src/miniprotocols/peersharing/client.rs
@@ -103,7 +103,7 @@ impl Client {
         let msg = Message::ShareRequest(amount);
         self.send_message(&msg).await?;
         self.0 = State::Busy(amount);
-        debug!("sent share request message with amount {}", amount);
+        debug!(amount, "sent share request message");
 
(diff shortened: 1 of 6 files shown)


pallas-network/src/miniprotocols/peersharing/client.rs
```diff
@@ -0,0 +1,134 @@
...
+        self.assert_inbound_state(&msg)?;
+
+        Ok(msg)
+    }
+
+    pub async fn send_share_request(&mut self, amount: Amount) -> Result<(), ClientError> {
+        let msg = Message::ShareRequest(amount);
+        self.send_message(&msg).await?;
+        self.0 = State::Busy(amount);
+        debug!("sent share request message with amount {}", amount);
```
scarmuega, line 106: no need for string interpolation in debug! traces. It's better to pass values as args to the macro.

franciscojoray: Thanks Santiago. Fixed it in the last commit.

scarmuega (approved):

Issue #296, opened 2023-10-03
Genesis config file parsing


scarmuega:
We want to be able to load genesis config files (byron, shelley, alonzo, etc) as defined by the Haskell node implementation in order to use those values in downstream components.

DoD:

- [ ] we can load & parse Byron genesis json files into structs
- [ ] we can load & parse Shelley genesis json files into structs
- [ ] we can load & parse Alonzo genesis json files into structs
- [ ] we can load & parse Conway? genesis json files into structs

rvcas:
Did I see this being started somewhere already? Or am I confused?

scarmuega:
> Did I see this being started somewhere already? Or am I confused?

We had #59 but I ended up constraining the scope to just Byron genesis because I needed a way to generate those initial UTxOs. Added this issue to keep track of work for the remaining config files.

SebastienGllmt:
There are still a few issues with genesis config parsing that affects the ability for downstream tools (ex: Dolos) to work with tools like yaci-devkit

You can find the result of the investigation by us & the devkit team here: https://github.com/bloxbean/yaci-devkit/issues/86#issuecomment-2547723603

PR #577, opened 2025-01-14
Fix Shelley genesis parsing

SebastienGllmt:
This fixes the Shelley part of #296

Notably,
- fix the type for `pools`
- fix the type for `initial_funds`

Additionally, this PR adds a new `shelley_utxos` helper function to help unblock this Dolos issue: https://github.com/txpipe/dolos/issues/427

## Background

This PR introduces a new "golden" test, which is taken from the Haskell codebase [here](https://github.com/IntersectMBO/cardano-ledger/blob/ff526dfe6460ae5a6ba398b091d6da52b539b285/eras/shelley/test-suite/test/Golden/ShelleyGenesis)

Notably, you can find the original code for this serialization logic in the Haskell codebase at the following:
- pool parsing [here](https://github.com/IntersectMBO/cardano-ledger/blob/ff526dfe6460ae5a6ba398b091d6da52b539b285/libs/cardano-ledger-core/src/Cardano/Ledger/PoolParams.hs#L231-L243)
- initial funds UTXO parsing [here](https://github.com/IntersectMBO/cardano-ledger/blob/ff526dfe6460ae5a6ba398b091d6da52b539b285/eras/shelley/impl/src/Cardano/Ledger/Shelley/Transition.hs#L447-L455)

### DIFF
--- a/pallas-configs/src/shelley.rs
+++ b/pallas-configs/src/shelley.rs
@@ -108,8 +108,72 @@ pub struct ProtocolParams {
 
 #[derive(Debug, Deserialize, Clone)]
 #[serde(rename_all = "camelCase")]
+pub struct Metadata {
+    pub hash: String,
+    pub url: String
+}
+
+#[derive(Debug, Deserialize, Clone)]
+pub struct SingleHostAddr {
+    pub port: Option<u32>,
+    pub IPv6: Option<String>,
+    pub IPv4: Option<String>,
+}
+
+#[derive(Debug, Deserialize, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct SingleHostName {
+    pub port: Option<u32>,
+    pub dns_name: String,
+}
+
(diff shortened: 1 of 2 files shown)


pallas-configs/src/shelley.rs
```diff
@@ -154,6 +233,25 @@ mod tests {
...
+        );
+        assert_eq!(
+            utxo.1.to_bech32().unwrap(),
+            "addr_test1qrsm4h32h9r95f8at64ykuugxqu3wvu0s5ay3vg6tlyevjh4e2flkegka00r69gt8c4vkxgf2vnnph3nsvhlkg5ukgxslee3tf"
+        );
+        assert_eq!(
+            utxo.2,
+            12157196
+        );
+    }
```
SebastienGllmt, line 253: I also tested this gives the same UTXO as running a yaci-devkit local node with this genesis block 👍

PR #584, opened 2025-01-22
feat(network): Bad inputs

sterraf:
This PR is based on #562 and introduces the `UtxoFailure` variant, which in turn includes `BadInputsUTxO`.
Encoding of errors is complete; for decoding, the auxiliary function `cbor_last` returns the raw data in case errors not yet identified.


sterraf:
Subsumed by #598

PR #590, opened 2025-02-04
feat(network): peer client version table query support

lancevincentsalera:
### Issue
- The handshake mini-protocol only replies with `MsgAcceptVersion` and `MsgRefuse` currently.
- Upon testing, the negotiated version doesn't inherit the actual PeerSharing config of the server
- So there was no actual way for us to know if the peer server has peersharing capabilities
- We needed a way for us to know the peer server's supported version table and version data
- We can only do that if the server replies with `QueryReply`
- ***However, if we merely add an option for us to do queries to the current setup, the connection closes as soon as a Query Reply is expected.***
- ***Thus, we need a separate function for it that connects only to perform a `query handshake`.***
- This specific feature is particularly useful for the peersharing mini-protocol implementation in Boros.

### Solution
- Create a query function inside peer client that connects to the given peer address and network magic
- Instead of subscribing to multiple channels, we only subscribe to handshake mini-protocol
- We negotiate with our own version table that has a `query = true`
- For our version table to support querying, we also need to update the version table to accept a query boolean param
- We return the supported version table from the peer server and we expect it to be a Query Reply.

### DIFF
--- a/pallas-network/src/facades.rs
+++ b/pallas-network/src/facades.rs
@@ -160,7 +160,7 @@ impl PeerClient {
         Ok(client)
     }
 
-    pub async fn query(
+    pub async fn handshake_query(
         addr: impl ToSocketAddrs,
         magic: u64,
     ) -> Result<VersionTable<VersionData>, Error> {
--- a/pallas-network/src/facades.rs
+++ b/pallas-network/src/facades.rs
@@ -173,6 +173,8 @@ impl PeerClient {
         let channel = plexer.subscribe_client(PROTOCOL_N2N_HANDSHAKE);
         let mut handshake = handshake::Client::new(channel);
 
+        let _plexer = plexer.spawn();
+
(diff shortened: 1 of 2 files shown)


Mercurial (dismissed):

Mercurial (approved):

pallas-network/src/facades.rs
```diff
@@ -159,6 +160,46 @@ impl PeerClient {
         Ok(client)
     }
 
+    pub async fn query(
```
scarmuega, line 163: the term `query` could be confused with local-state-queries. Lets use `handshake_query`.

scarmuega (dismissed):

pallas-network/src/facades.rs
lancevincentsalera: Thanks for catching that, @scarmuega . Already addressed your comment in the latest commit. Let me know if there’s anything else!

Mercurial (approved):

scarmuega (approved):

PR #588, opened 2025-02-03
feat(network): Collateral-related failures

sterraf:
This PR (dependent on #585) implements five of Conway's `UtxoFailure` variants related to collaterals.

Unit-tests derived from node interactions are provided for the variants `InsufficientCollateral` and `NoCollateralInputs`.

PR #593, opened 2025-02-06
feat(network): Conway's UtxowPredFailure variants

sterraf:
This PR (based on #588) implements three more variantes of the `UtxowFailure` enum, namely:

+    `PPViewHashesDontMatch`,
+    `NotAllowedSupplementalDatums`, and
+    `ExtraRedeemers`.

(Unit-tests for the first and third variant were included before.)

It also implements the `SMaybe` option type with the codec corresponding to (`S`)`Maybe` type of the Cardano node, required for the first variant above.


sterraf:
Subsumed by #598

PR #594, opened 2025-02-06
Change PoolMetadata.hash type to Bytes

ginnun:
In the `ledger` code base, `PoolMetadata`'s hash field is defined as a `ByteString` of any length. Although at runtime it is expected to get 32 bytes long hash, the datatype is not limited to it. We are getting empty hashes while using the `Arbitrary` instances to test our Rust code. This causes CBOR decoding to fail.

We created a PR against the `ledger` codebase but It got rejected [here](https://github.com/IntersectMBO/cardano-ledger/pull/4870#pullrequestreview-2597267579)

So here the PR for changing the field to `Bytes`. This change is not a breaking one.

cc @michalrus


michalrus (approved):
Agreed, this field can be a byte array of _any_ length. ✅

scarmuega:
I think this has been already introduced by #609

PR #596, opened 2025-02-10
Add unknown cost models into the CostModels type

ginnun:
Although not likely to receive such a CBOR, we want to be able to handle decoding unknown cost models when it happens.


michalrus (approved):
LGTM :+1:

PR #585, opened 2025-01-27
feat(network): Conway UTxO substransition failures

sterraf:
This PR is based on #584 and introduces the (Conway)`UtxosPredFailure` subtransition failure enum and two of its variants. Unit-tests using inputs from a real interaction with a node are provided.

PR #598, opened 2025-02-12
feat(network): Local Tx Submission errors

sterraf:
This PR decodes the `RejectReason` after a failed local Tx submission by structuring into two variants:

1. Plutus errors (an ASCII string); or
2. an `(Era, Vec<TxError>)` pair.

The `TxError` type corresponds to the Conway era ledger transaction errors. Its only analyzed variant is that of [ConwayUtxowPredFailure](https://github.com/IntersectMBO/cardano-ledger/blob/d30a7ae828e802e98277c82e278e570955afc273/eras/conway/impl/src/Cardano/Ledger/Conway/Rules/Utxow.hs#L94) for which, in turn, 6 variants have been implemented along with their deeper structure. 

For each error having variants not yet analyzed, a temporary "raw" variant was added to hold the CBOR information.

Unit-tests using an up-to-date preprod node response were added to `codec.rs`. For instance, I had to implement the `SMaybe` option type with the codec corresponding to (`S`)`Maybe` type of the Cardano node, required for many types (different to default one for `Option<T>`.

This PR properly subsumes #562, #584, #585, #588, and #593, which are to be closed.


scarmuega (approved):

ginnun:
For anyone interested, we have a full working implementation of all the tx submission errors in this fork: https://github.com/blockfrost/pallas/tree/feat/blockfrost

We plan to open a PR once we do a bit of refactoring.

PR #607, opened 2025-02-19
feat: pallas validate crate implementation 1st pass

Mercurial:


pallas-validate/Cargo.toml
```diff
@@ -22,6 +13,12 @@ rand = "0.8"
...
 thiserror = "1.0.49"
+serde = { version = "1.0.136", features = ["derive"] }
+serde_json = "1.0.122"
 
-[dev-dependencies]
-hex = "0.4"
+# phase2 dependencies
+uplc = { git = "https://github.com/pragma-org/uplc.git" }
+rug = { version = "1.26.1" }
+miette = { version = "7.5.0", features = ["fancy"] }
```
scarmuega, line 22: do we rellay need miette here? what for?

Mercurial:
We discovered a bug that pallas is overriding the redeemer structure into array instead of leaving it alone. in Conway redeemers have now two types `array` and `map` so if a tx uses the `map` and pallas overrides the `map` into `array` then the `script_data_hash` changes

scarmuega:
Looks good, just two comments:
- remove duplicated entry on workspace
- add `phase2` feature flag to pallas-validate and pallas

scarmuega (approved):

Issue #605, opened 2025-02-18
refactor(primitvies): remove roundtrip-safe structs from primitives


scarmuega:

sterraf:
It seems necessary to implement the corresponding `serde` traits for `KeepRaw`, in order to have the same functionality. Should they be "transparent", i.e., ignoring the CBOR data? Or the derived ones for `KeepRaw` work well?

[Today](https://github.com/txpipe/pallas/blob/4636b2d32a7cb126690495b5bd96cfa29763f882/pallas-primitives/src/alonzo/model.rs#L895-L911), derived traits are not the same for `Minted` vs plain structs: 
```rust
#[derive(Serialize, Deserialize, Encode, Decode, Debug)]
pub struct Tx {
    ...
```
vs
```rust
#[derive(Encode, Decode, Debug, Clone)]
pub struct MintedTx<'b> {
    ...
```

scarmuega:
@sterraf `KeepRaw` should be transparent for serde.

sterraf:
[There are instances](https://github.com/txpipe/pallas/blob/main/pallas-primitives/src/alonzo/model.rs#L547) of roundtrip-safe structs that have to be moved inside of a `KeepRaw`, namely:
```rust
pub struct MintedBlock<'b> {
    #[n(0)]
    pub header: KeepRaw<'b, MintedHeader<'b>>,

    #[b(1)]
    pub transaction_bodies: MaybeIndefArray<KeepRaw<'b, TransactionBody>>,
```
In order to safely switch to `Vec`, it would be natural to do:
```rust
    pub transaction_bodies: MaybeIndefArray<KeepRaw<'b, TransactionBody>>,
```
This motivates: *Should `Minted` structs (which are going to be default by #604) have a more uniform layout, as in "each field is a `KeepRaw`?*

PR #611, opened 2025-02-26
refactor: Eliminating roundtrip-codec safe structures

sterraf:
WIP addressing #605.

The changes were propagate through `traverse`, and they work included tests at `hardano` and `config`.

There remain to be fixed (at least) `txbuilder` and `applying`.

### DIFF
--- a/pallas-primitives/src/alonzo/model.rs
+++ b/pallas-primitives/src/alonzo/model.rs
@@ -11,11 +11,13 @@ pub use pallas_codec::codec_by_datatype;
 pub use crate::{
     plutus_data::*, AddrKeyhash, AssetName, Bytes, Coin, CostModel, DatumHash, DnsName, Epoch,
     ExUnitPrices, ExUnits, GenesisDelegateHash, Genesishash, Hash, IPv4, IPv6, Int, KeepRaw,
-    KeyValuePairs, MaybeIndefArray, Metadata, Metadatum, MetadatumLabel, NetworkId, Nonce,
-    NonceVariant, Nullable, PlutusScript, PolicyId, PoolKeyhash, PoolMetadata, PoolMetadataHash,
+    Metadata, Metadatum, MetadatumLabel, NetworkId, Nonce,
+    NonceVariant, PlutusScript, PolicyId, PoolKeyhash, PoolMetadata, PoolMetadataHash,
     Port, PositiveInterval, ProtocolVersion, RationalNumber, Relay, RewardAccount, ScriptHash,
     StakeCredential, TransactionIndex, TransactionInput, UnitInterval, VrfCert, VrfKeyhash,
 };
+
+use crate::BTreeMap;
 
 #[derive(Serialize, Deserialize, Encode, Decode, Debug, PartialEq, Eq, Clone)]
 pub struct HeaderBody {
@@ -96,7 +98,7 @@ impl<'a> From<MintedHeaderBody<'a>> for HeaderBody {
     }
 }
 
-pub type Multiasset<A> = KeyValuePairs<PolicyId, KeyValuePairs<AssetName, A>>;
+pub type Multiasset<A> = BTreeMap<PolicyId, BTreeMap<AssetName, A>>;
 
 pub type Mint = Multiasset<i64>;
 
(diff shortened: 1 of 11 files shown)


pallas-primitives/src/alonzo/model.rs
```diff
@@ -142,7 +144,7 @@ pub enum InstantaneousRewardSource {
 
 #[derive(Serialize, Deserialize, Debug, PartialEq, Eq, PartialOrd, Ord, Clone)]
 pub enum InstantaneousRewardTarget {
-    StakeCredentials(KeyValuePairs<StakeCredential, i64>),
+    StakeCredentials(BTreeMap<StakeCredential, i64>),
     OtherAccountingPot(Coin),
```
sterraf, line 148: It seems that this `KeyValuePairs` (at least) was really a proxy for literally an array of pairs (that encodes as such), and not really a map.

Is this correct? Should I used such an array here instead of `BTreeMap`?

pallas-primitives/src/alonzo/model.rs
```diff
@@ -612,7 +610,7 @@ pub struct MintedTx<'b> {
     pub success: bool,
 
     #[n(3)]
-    pub auxiliary_data: Nullable<KeepRaw<'b, AuxiliaryData>>,
+    pub auxiliary_data: KeepRaw<'b, Option<AuxiliaryData>>,
 }
```
sterraf, line 614: This was a tough one, since if I replace `Nullable` by `Option` in place, it turns into a "optional field"  which is not encoded at all if nil.

I opted for having the `Option` inside `KeepRaw` in order to preserve the cbor.

pallas-primitives/src/alonzo/model.rs
```diff
@@ -636,7 +634,7 @@ mod tests {
             include_str!("../../../test_data/alonzo6.block"),
             include_str!("../../../test_data/alonzo7.block"),
             include_str!("../../../test_data/alonzo8.block"),
-            include_str!("../../../test_data/alonzo9.block"),
+            // include_str!("../../../test_data/alonzo9.block"),
             // old block without invalid_transactions fields
```
sterraf, line 638: I kept the block-isomorphism test while commenting out the two offending ones.

pallas-applying/src/utils.rs
```diff
@@ -28,17 +28,9 @@ pub use validation::*;
...
-        }
-        _ => {
-            (mtx.transaction_body.raw_cbor().len() + mtx.transaction_witness_set.raw_cbor().len())
-                as u32
-        }
-    }
+    (mtx.auxiliary_data.raw_cbor().len()
+     + mtx.transaction_body.raw_cbor().len()
+     + mtx.transaction_witness_set.raw_cbor().len()) as u32
 }
```
sterraf, line 34: Now that we have the `KeepRaw` encapsulating the whole field, this simpler formula should do the work, afaik.

pallas-applying/src/utils.rs
```diff
@@ -659,15 +639,27 @@ pub fn is_byron_address(address: &[u8]) -> bool {
...
 
 pub fn aux_data_from_alonzo_minted_tx<'a>(mtx: &'a AlonzoMintedTx) -> Option<&'a [u8]> {
-    mtx.auxiliary_data.as_ref().map(|x| x.raw_cbor()).into()
+    if let Some(_) = mtx.auxiliary_data.deref() {
+        Some(mtx.auxiliary_data.raw_cbor())
+    } else {
+        None
+    }
 }
```
sterraf, line 647: This is a bit verbose, I'll try a slicker form afterwards (same repeated two times below in the same file)

sterraf:
@scarmuega There are some comments on the more nuanced decisions above. Please tell me if this is the right direction in order to keep up with the updated `applying` and to propagate to the rest of the codebase (`u5c`, for instance).

pallas-validate/src/phase_one/shelley_ma.rs
```diff
@@ -279,13 +283,14 @@ fn check_metadata(tx_body: &TransactionBody, mtx: &MintedTx) -> ValidationResult
...
-    tx_body: &TransactionBody,
+    tx_body: &KeepRaw<'_, TransactionBody>,
     tx_wits: &MintedWitnessSet,
     utxos: &UTxOs,
 ) -> ValidationResult {
     let vk_wits: &mut Vec<(bool, VKeyWitness)> =
         &mut mk_alonzo_vk_wits_check_list(&tx_wits.vkeywitness, ShelleyMA(MissingVKWitness))?;
-    let tx_hash: &Vec<u8> = &Vec::from(tx_body.compute_hash().as_ref());
+    let tx_hash: &Vec<u8> = &Vec::from(tx_body.original_hash().as_ref());
+    println!("Tx hash: {:02x?}", tx_hash.clone());
```
scarmuega, line 293: is this required?

sterraf: The second line is not needed, thanks. It was there for debugging.

scarmuega (approved):

PR #624, opened 2025-03-07
Lots of fixes and updated to the latest node version

ginnun:
This PR contains lots of fixes, improvements and fixes some breaking changes that comes with the latest Cardano node version (for Haskell text generation)

## Summary by CodeRabbit

- **New Features**
  - Structured reward-account format with on-wire encode/decode.
  - Optional Plutus V4 cost model and PlutusV4 language tagging.
  - Reusable mismatch container type and new error variants for richer diagnostics.

- **Changes**
  - Many protocol types now support ordering.
  - Transaction decoding tightened; numeric handling expanded (larger integers).
  - Error/serialization APIs now return fallible results.

- **Tests**
  - Updated to cover Plutus V4, new account format, mismatch handling, and revised decoding behavior.


pallas-network/src/miniprotocols/localstate/queries_v16/mod.rs
```diff
@@ -491,16 +491,41 @@ pub enum Vote {
     Abstain,
 }
 
-pub type RewardAccount = Bytes;
+#[derive(Debug,  Clone, Eq, PartialEq, PartialOrd, Ord)]
+pub struct FieldedRewardAccount{
```
scarmuega, line 495: is the `FieldedRewardAccount` struct defined as part of some spec or is it a quality-of-life improvement for dealing with credentials?

If its the latter, I would prefer to have this in a higher layer of the stack and not as part of the codec of the miniprotocol.

ginnun: @scarmuega It's there to be able to produce the same error message with Haskell codebase. Since Rust doesn't have type specialization, old-good RewardAccount (which is a Byte) can't be  used for HaskellDisplay to distinguish. 

So not sure if this is a quality-of-life improvement or not. If do you think so, I could refactor it. Just tell me what layer it should be in that case. Thanks.

michalrus (commented):
Couple of comments, since we’re depending on this branch, and it hasn’t been been approved yet:

michalrus (approved):
The CBOR tests are passing in:
- https://github.com/blockfrost/blockfrost-platform/pull/514

pallas-network/src/miniprotocols/localstate/queries_v16/mod.rs
```diff
@@ -557,16 +629,82 @@ pub enum Vote {
...
+        let header_type = bytes[0] >> 4;
+        if !matches!(header_type, 0b1110 | 0b1111) {
+            return Err(InvalidRewardAccount::InvalidHeaderType(bytes[0]));
+        }
+
+        let network = if bytes[0] & 0b0000_0001 != 0 {
+            Network::Mainnet
+        } else {
+            Network::Testnet
+        };
```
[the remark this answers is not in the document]
michalrus: @ginnun WDYT?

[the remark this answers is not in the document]
ginnun: @michalrus ⚙️ cfa0c63

michalrus (approved):
LGTM!

pallas-primitives/src/conway/model.rs
```diff
@@ -725,6 +731,37 @@ pub struct Tx<'b> {
 
 impl Eq for Tx<'_> {}
 
+impl<'b, C> minicbor::Decode<'b, C> for Tx<'b>
```
scarmuega: Could you walk me through what motivated this custom `Decode` for `Tx<'b>`?

```rust
let success = match d.decode_with(ctx) {
    Ok(v) => v,
    Err(e) if e.is_end_of_input() => true,
    Err(e) => return Err(e),
};
let auxiliary_data = match d.decode_with(ctx) {
    Ok(v) => v,
    Err(e) if e.is_end_of_input() => Nullable::Null,
    Err(e) => return Err(e),
};
```

is it required at the CDDL level for conway?

ginnun: Good catch, you're right, the Conway CDDL doesn't require this.

The original idea behind the manual impl was tolerance for pre-Alonzo tx layouts ([body, witness_set, auxiliary_data / nil] and no isValid)

Decoding should fail if it's not a four-element array.

ginnun: Applied in  9a83c74125738d198010dcb4d61a3595a6d4d771

michalrus:
@scarmuega, hey. 👋 I've just pushed some updates for Cardano Node 11.0.1 (van Rossem / PV11) – without them almost all of our random tests fail. After this commit, there are no failures.

Issue #618, opened 2025-03-06
Remove Pseudo struct from primitives


scarmuega:
We need to remove all `PseudoXXX` structs from pallas-primitives. These structs are not longer needed since we're unifying minted vs non-minted versions of the ledger structs.

PR #632, opened 2025-03-25
feat(primitives): Remove Babbage & Conway Pseudo structs

sterraf:
WIP addressing #618 for the rest of eras.

Comment to be expanded afterwards.


sterraf:
@scarmuega There are some places (notably `pallas-txbuilder`) where a `KeepRaw` value must be constructed. I believe the only way to do this is to decode some `u8` slice, and thus lifetime issues arise.

The last commit ([2236b71](https://github.com/txpipe/pallas/pull/632/commits/2236b71cadbc3720c60856fc6111ed71863d4199)) shows how this was handled for the `pallas-validate` test suite, by reserving some `Vec<u8>` for the purpose of having references that live long enough.

Is this the correct workaround?

sterraf:
Two tests in the `phase2` suite at `pallas-validate` require a faithful CBOR representation at the `KeepRaw` values, hence the [last commit](https://github.com/txpipe/pallas/pull/632/commits/9b9bc4c8ab74401f266532da8b95c866a7e46e33) adds the corresponding codec-safe auxiliary functions.

PR #633, opened 2025-03-26
refactor(validate): uplc codebase based on pallas-primitives changes

lancevincentsalera:


Mercurial (approved):

pallas-validate/src/uplc/script_context.rs
```diff
@@ -762,7 +760,8 @@ fn script_purpose_builder<'a>(
         match tag {
             RedeemerTag::Mint => mint
                 .mint_value
-                .get(index)
+                .iter()
+                .nth(index)
```
sterraf, line 764: I'm a bit concerned about the use of `nth` here and below (comparing to the behavior prior to the elimination of `KeyValuePairs`), because you are referring to the mint key-value by its inherited order (and now not exactly the CBOR order).

It might not be a real problem though, perhaps @scarmuega can tell.

sterraf (commented):
I only left one comment on the possibly different order of `BTreeMap` (as part of a `Multiasset`) vs that of the former `KeyValuePairs`.

pallas-validate/src/uplc/script_context.rs
scarmuega: the index relies on the mint items to be ordered lexicographically by policy hash IIRC. That sort should be applied here explicitly before looking for the nth value.

sterraf: I understand that this is exactly the order for the `BTreeMap` keys.

scarmuega (approved):

PR #587, opened 2025-01-30
Cardano blueprint handshake tests

ch1bo:
This demonstrates how we could test against test vectors in a common repository from multiple implementations of the ouroboros protocols.

For that matter, simple serialization roundtrip tests against the handshake protocol cddl are added in this PR.

The `cardano-blueprint` is added as a submodule, but tests using data from it are gated by a `blueprint` feature flag.

This means, new contributors would experience this workflow:
- `git clone git@github.com:txpipe/pallas.git && cd pallas`
- `cargo test` -> works, while tests using `cardano-blueprint` are not run
  - Contributor starts hacking on things
  - NOTE: Any changes are checked in CI _including the `blueprint` tests_!
- After a CI failure or when deliberately wanting to run the blueprint tests:
  - `cargo test --features blueprint` -> fails because submodule not checked out
  - `git submodule update --init` checks out submodule
  - `cargo test --features blueprint` -> works now

PR #638, opened 2025-04-01
Cardano blueprint localstatequery tests

ch1bo:
Stacked with #587, review that first.

Explores how we can test the [LocalStateQuery API](https://github.com/cardano-scaling/cardano-blueprint/issues/25) that is being created in course of the blueprint initiative.

~~Note that the first (and only) API query example for `getSystemStart` is actually failing to decode the result, because the `year` is actually a big negative number = the rust implementation behaves different than the CDDL specification (I personally think unsigned numbers would be in fact nicer though .. or maybe even use `time` for this particular query result).~~
This PR now also fixes the `SystemStart` data type (incl. derived encoder + decoder)

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -41,6 +41,8 @@ jobs:
     steps:
       - name: Checkout sources
         uses: actions/checkout@v4
+        with:
+          submodules: true
 
       - name: Install stable toolchain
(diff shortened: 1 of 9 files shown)


scarmuega (approved):

scarmuega:
and so it begins!

PR #649, opened 2025-04-16
Deprecate pallas wallet

### LINKED ISSUE #619: Deprecate pallas-wallet crate
Pallas-wallet is an opinionated implementation of an encrypted storage for private keys. Given the newly defined scope for Pallas (only protocol-level stuff, not concrete implementations), this crate no longer belongs here.

Steps for deprecation:

- [ ] Move some very basic version of this logic into a new example in the examples folder. Goal should be to show how to manage private keys in a secure way.
- [ ] Copy relevant code to known users of the crate. In particular: [CShell](https://github.com/txpipe/cshell)

logicalmechanism:
With the introduction of #647, we should be able to close #619.

- `pallas-wallet` is now in the example folder.
- All `pallas-wallet` dependencies should be removed.
- Readme was updated.

There will probably be some minor fixes to the readme or the example folder, but this should at least remove `pallas-wallet` for now.


scarmuega (approved):

scarmuega:
perfect!!

PR #661, opened 2025-06-03
fix: cost models into vec lexicographically

nicolasLuduena:
Values of cost models are supposed to be sorted in lexicographical manner. Consequences could be seen on the blaze provider of utxorpc  that got all cost models mixed up.

### DIFF
--- a/pallas-configs/src/alonzo.rs
+++ b/pallas-configs/src/alonzo.rs
@@ -77,7 +77,9 @@ pub struct CostModel(HashMap<String, i64>);
 
 impl From<CostModel> for Vec<i64> {
     fn from(value: CostModel) -> Self {
-        value.0.into_values().collect()
+        let mut entries: Vec<_> = value.0.into_iter().collect();
+        entries.sort_by_key(|(k, _)| k.clone());
+        entries.into_iter().map(|(_, v)| v).collect()
     }
 }
 


Quantumplation:
I believe they're *supposed* to be, but that's convention based on the serializer in the haskell node, and in some eras has not followed that convention. This has burned me in the past.

I'm not 100% sure that's true, so we should check it carefully in all eras; if I'm right, then enforcing this in pallas itself will break projects that need to be able to sync previous eras (like archive nodes etc.)

Quantumplation:
Before this PR, the cost model was sorted, not arbitrarily, but in the order it was constructed / deserialized.

After this, it sorts them lexographically.

In the current era, the lexographical behavior is correct.

In other eras, I believe this wasn't enforced, and it was the order the fields appeared in the Haskell record.

So, if you make this change, and someone consuming this code was taking care to construct the cost models in the correct order, you'll now be changing that behavior.

In particular this matters for the script data hash calculation, and so you'd break validating the ledger in old eras.

Again, I can't point you to a source for this, as it's just based on some frustrating debugging sessions from a year or more ago, and I could be mistaken, but I'm just raising it because you'll want to make sure this doesn't break anything. And if it does, I would recommend doing this sorting with different era specific constructors, rather than when you grab the cost model values (which might be better for performance anyway)

scarmuega:
Hi @nicolasLuduena — thanks for catching this and for the patch, and apologies for the slow turnaround.

Since this PR was opened, `CostModel` in `pallas-configs` has been restructured in a way that addresses the underlying issue:

- The internal representation changed from `CostModel(HashMap<String, i64>)` to `CostModel(Language, Vec<i64>)`. There's no longer a `HashMap` to iterate, so the unstable-order bug that caused the scrambled output in blaze can't occur.
- The `Vec<i64>` is populated in canonical order at construction time in `from_array_with_language` / `from_object_with_language`, using `cost_models::get_names_for_version(plutus_version)` as the source of truth. `From<CostModel> for Vec<i64>` now simply returns the inner vec.
- This also addresses the concern @Quantumplation raised in the thread: ordering is per-language rather than a global lex sort, so older eras can keep their own canonical order if it ever diverges from alphabetical.

So the fix is no longer applicable as-is — the patch site doesn't exist anymore, and the behavior you wanted is what the code does today. Closing for housekeeping. If you spot any case where the cost-model order still comes out wrong on current `main`, please open a fresh issue and we'll dig in. Thanks again for the contribution!

PR #662, opened 2025-06-11
chore: replace dashu math library with num-bigint in pallas-math

whankinsiv:
This PR removes the dashu math library dependency and replaces it with num-bigint to standardize math operations across the codebase, improving consistency.

### DIFF
--- a/pallas-math/Cargo.toml
+++ b/pallas-math/Cargo.toml
@@ -12,7 +12,6 @@ authors = ["Andrew Westberg <andrewwestberg@gmail.com>"]
 exclude = ["tests/data/*"]
 
 [dependencies]
-bigdecimal = "0.3"
 num-integer = "0.1"
 num-bigint = "0.4"
 num-traits = "0.2"
--- a/pallas-math/Cargo.toml
+++ b/pallas-math/Cargo.toml
@@ -12,8 +12,10 @@ authors = ["Andrew Westberg <andrewwestberg@gmail.com>"]
 exclude = ["tests/data/*"]
 
 [dependencies]
-dashu-int = "0.4.1"
-dashu-base = "0.4.1"
+bigdecimal = "0.3"
+num-integer = "0.1"
+num-bigint = "0.4"
+num-traits = "0.2"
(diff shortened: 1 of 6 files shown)


AndrewWestberg:
Did you test the performance implications of this? Num-bigint was slow to validate blocks when I tested various math libraries back in the day.

whankinsiv:
> Did you test the performance implications of this? Num-bigint was slow to validate blocks when I tested various math libraries back in the day.

Hi Andrew. Here are my benchmark results for each operation used is math_dashu.rs:
Screenshot 2025-06-13 at 2 06 31 PM
If you’d like to dive deeper into the benchmark data, I’d be happy to share a Dropbox link or any other format that works for you.

KtorZ:
I would prefer sticking to one library; and I don't think the performances should be the sole argument. Both libraries have very decent perfs and I hardly think that any will constitute a bottleneck in block header validations.

Also, rather than comparing rust libs against one another in terms of perfs, we should compare rust lib with the current Haskell implementation. Because if they're all slower, then they all equally suck :). 

But what are the other factors? 

- interoperability / cross compilation: and on that regard, since we already cross compile our header validation code just fine with dashu and the ledger with num-bigint; then it seems fine with both.

- maintenance / robustness: num-bigint is more established and with about half a million of daily downloads; vs 2000 downloads for dashu. So the chances that something is off in dashu vs num-bigint is unfortunately higher. Although both seems well tested; the risk with dashu is higher. 

- module API: I have to say that I prefer num-bigint's API so far. But that one is purely subjective and likely due to habits. 


One way for way with Pallas could be to enable either depending on feature flags; although I think Pallas already went down that path and came back? So perhaps there's no interest in supporting multiple maths libraries.

Quantumplation:
I think all of that is sensible.

It might be worth writing a benchmark specifically for header validation, and just establish a reference budget using the haskell node numbers.

AndrewWestberg:
The golden tests in the haskell code are a good benchmark to hit. The reference C code that haskell is matching used gmp which is used by rug.

PR #670, opened 2025-07-05
fix: encoding and decoding of NativeScript

KtorZ:
There were 2 problems here:

  1. The encoder would produce invalid CBOR in for the N-of-K variant of NativeScript which is encoded as an array of 3 elements, but marked as containing 2.

  2. The decoder doesn't consume all the bytes it should; thus succeeding to decode partial / incomplete (and thus invalid) CBOR sequences. Note that while fixed in this particular instance for NativeScript, the problem is recurring across pretty much all list-like decoders within Pallas and should probably be subject to a larger overhaul.

### DIFF
--- a/pallas-primitives/src/alonzo/model.rs
+++ b/pallas-primitives/src/alonzo/model.rs
@@ -570,23 +570,61 @@ pub enum NativeScript {
 
 impl<'b, C> minicbor::decode::Decode<'b, C> for NativeScript {
     fn decode(d: &mut minicbor::Decoder<'b>, ctx: &mut C) -> Result<Self, minicbor::decode::Error> {
-        d.array()?;
+        let size = d.array()?;
+
+        let assert_size = |expected| {
+            // NOTE: unwrap_or allows for indefinite arrays.
+            if expected != size.unwrap_or(expected) {
+                return Err(minicbor::decode::Error::message(
+                    "unexpected array size in NativeScript",
+                ));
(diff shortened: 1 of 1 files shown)


pallas-primitives/src/alonzo/model.rs
```diff
@@ -596,31 +634,35 @@ impl<C> minicbor::encode::Encode<C> for NativeScript {
...
                 e.encode_with(1, ctx)?;
                 e.encode_with(v, ctx)?;
             }
             NativeScript::ScriptAny(v) => {
+                e.array(2)?;
                 e.encode_with(2, ctx)?;
                 e.encode_with(v, ctx)?;
             }
             NativeScript::ScriptNOfK(a, b) => {
+                e.array(3)?;
```
KtorZ, line 654: ☝️

pallas-primitives/src/alonzo/model.rs
```diff
@@ -570,23 +570,61 @@ pub enum NativeScript {
...
             _ => Err(minicbor::decode::Error::message(
                 "unknown variant id for native script",
             )),
+        }?;
+
+        if size.is_none() {
+            let next = d.datatype()?;
+            if next != minicbor::data::Type::Break {
+                return Err(minicbor::decode::Error::type_mismatch(next));
+            }
```
KtorZ, line 624: ☝️

PR #685, opened 2025-08-11
feat: Make Pallas WASM compatible

bkioshn:
# Description

We are currently implementing [Hermes](https://github.com/input-output-hk/hermes) WebAssembly (WASM) application engine designed to provide secure, sandboxed execution of modular applications. Because Hermes targets the WASM platform, it inherits WASM’s inherent limitations - that WASM modules cannot directly access operating system features such as I/O, networking, or threading.

The primary obstacle preventing the WASM module from importing Pallas (either directly or through transitive dependencies) is the `pallas-network` crate. This crate relies on Tokio and socket APIs for networking, which are incompatible with the WASM environment. Meanwhile, Hermes requires certain constants and general-purpose types currently defined within `pallas-network`. Attempting to import `pallas-network` into Hermes results in build failures due to these I/O and async runtime dependencies.

To resolve this, we propose extracting these constants and primitive types into a existing crate, `pallas-primitive`, which has no dependencies on Tokio or socket APIs. This restructuring allows Hermes to use the necessary shared definitions without bringing in incompatible code. Additionally, this PR includes cleanup of redundant types across the Pallas crates, improving overall code clarity and maintainability. 

In addiiton, `pallas` itself, when import directly should be WASM compatible by disable some feature like `pallas-hardano` and `pallas-network` for `wasm` family architect.

# Summary of changes
- Extract `Point` and constant out of `pallas-network` to `pallas-primitives`. Note the old one as deprecated
- Remove redundant constant in `pallas-traverse`
- Disable `pallas-network` and `pallas-hardano` when architect family is `wasm`

## Summary by CodeRabbit

* **New Features**
  * Introduces centralized network constants and a canonical Point type for consistent use across the ecosystem.

* **Chores**
  * Legacy constants and Point locations deprecated and re-exported — update imports to the new canonical paths.
  * Network and Hardano-related exports now gated away from wasm targets to improve cross-target compatibility.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


scarmuega:
Thanks @bkioshn, I agree with that this is something we need to support.

The `Point` struct should be moved to primitives, but the I'm not so sure about the network constants. Those are pretty coupled with code that actually does networking.

What would be the use case of having those constants available in a environment that doesn't support networking?

bkioshn:
> Thanks @bkioshn, I agree with that this is something we need to support.
> 
> The `Point` struct should be moved to primitives, but the I'm not so sure about the network constants. Those are pretty coupled with code that actually does networking.
> 
> What would be the use case of having those constants available in a environment that doesn't support networking?

We are using those constant network in our cardano-blockchain-types
https://github.com/input-output-hk/catalyst-libs/blob/main/rust/cardano-blockchain-types/src/network.rs
This crate is design to be WASM compatible, so if this is WASM compatible, it cannot use the `pallas-network` directly

PR #703, opened 2025-09-14
fix(configs): avoid weird ratios in config float parsing

scarmuega:


rodrigosuaya:
@coderabbitai full review

PR #704, opened 2025-09-15
Remove kes cli crate

paweljakubas:
As `clap` is the source of problems for libraries that use `pallas` and that it might be not the best place to have CLI app, this PR removes `kes-cli` along with the traces that point to it


paweljakubas:
FYI @scarmuega

AndrewWestberg (approved):

PR #707, opened 2025-10-15
implement pedantic codec validations via minicbor context

rrruko:
This keeps the default permissive behavior of the codecs while providing the option to enable pedantically node-conformant validations by using `minicbor::decode_with(..., BasicStrictContext::new())`.

See also https://github.com/txpipe/pallas/pull/697, https://github.com/pragma-org/amaru/pull/494

## Summary by CodeRabbit

- New Features
  - Stricter, schema-validated decoding for values and transaction bodies, rejecting zero or empty amounts and duplicate/invalid fields.
  - Non-empty collections for multi-asset, mint and withdrawals to ensure meaningful data and clearer validation errors.
  - Field-based transaction body validation with helpers to build and validate core transaction fields.

- Refactor
  - Mint handling and value encoding/decoding rewritten to enforce non-empty, non-zero assets and improve robustness.

### DIFF
--- a/pallas-primitives/src/conway/model.rs
+++ b/pallas-primitives/src/conway/model.rs
@@ -814,6 +814,102 @@ mod tests {
         }
     }
 
+    mod tests_witness_set {
+        use super::super::{Bytes, VKeyWitness, WitnessSet};
+        use pallas_codec::minicbor;
+
+        #[test]
+        fn decode_empty_witness_set() {
+            let witness_set_bytes = hex::decode("a0").unwrap();
+            let ws: WitnessSet = minicbor::decode(&witness_set_bytes).unwrap();
+            assert_eq!(ws.vkeywitness, None);
+        }
+
+        #[test]
+        fn decode_witness_set_having_vkeywitness_legacy_may_be_empty() {
+            let witness_set_bytes = hex::decode("a10080").unwrap();
+            let ws: WitnessSet = minicbor::decode(&witness_set_bytes).unwrap();
+
+            // TODO: It is really surprising that the guard when decoding non
+            // empty sets from cbor has been commented out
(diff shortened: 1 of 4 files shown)


pallas-primitives/src/conway/model.rs
```diff
@@ -27,25 +27,216 @@ pub use crate::babbage::OperationalCert;
 
 pub use crate::babbage::Header;
 
-pub type Multiasset<A> = BTreeMap<PolicyId, BTreeMap<AssetName, A>>;
+use pallas_codec::minicbor::data::Type;
 
-pub type Mint = Multiasset<NonZeroInt>;
+trait StrictContext {
```
Quantumplation: Some suggestions:
 - If we name this something more general, like `ValidationContext`,
 - and have the push_error should return a Result that we can `?`;

Then we can have a NoopContext, which does nothing; an AccumulatingContext which accumulates things as errors, and a TerminatingContext, which returns an error.

Then, our `Strict<T>` implements minicbor::Decode, constructs a `TerminatingContext`, and deserializes T;

And we could have a `StrictVerbose<T>` that implements `minicbor::Decode`, constructs an `AccumulatingContext`, and then concatenates all the errors to return it's own at the end.

yHSJ (commented):
I feel like I'd have to try out this API to really get an opinion on whether or not I love it.

My immediate concerns are:
1) If we `unwrap` a Strict wrapper (I don't like using that function name, feels dirty to me), I lose any "proof" that I can trust the data inside it instead of revalidating stuff.
2) Currently, the pallas decoders fail sometimes (when the data is in the wrong format, for example, but not "business logic" validations). How would this change differentiate between business logic and fundamental issues with the data?

But really, I think maybe we should just give it a try and see how it feels as an API.


*Note: I have not reviewed the actual logic of the decoders, I am simply commenting on the usability of the API*

cc: @scarmuega @KtorZ

PR #719, opened 2025-12-13
Add VRF support and update documentation

DaJo-Code:
Introduce support for Verifiable Random Functions (VRF) alongside KES, including necessary documentation updates. This enhancement allows users to generate proofs and verify outputs for both draft-03 and draft-13 variants of VRF.

## Summary by CodeRabbit

* **New Features**
  * Added VRF (Verifiable Random Function) support with draft-03 and draft-13 implementations, available as an optional feature.

* **Documentation**
  * Updated documentation to describe VRF support and experimental status.

* **Tests**
  * Added comprehensive VRF test coverage including IETF vector validation and KES regression tests for deterministic key generation and signature verification across periods.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


scarmuega:
@FractionEstate your work on the Cardano crypto repo / crate looks very thorough 👏 

My concern with this PR is that it doesn't align with Pallas scope. We try to be as lightweight as possible in terms of dependencies (not always easy, but we try).

The scope of Pallas is being a bag of independent building blocks that can co-exist with other implementations (like yours). I believe the decision of which blocks to compose together should be made by the final user instead of Pallas.

If you agree, what I would do in this scenario is add a reference in the readme explaining that for vrf support there's cardano-crypto crate available.

PR #792, opened 2026-07-20
feat(network): add local-tx-monitor server agent, align codec with the wire protocol

Javieracost:
## What

Server-side support for the local-tx-monitor miniprotocol (N2C protocol 9), plus the codec fixes required for it to interoperate with real peers. Three commits, each self-contained:

1. `chore: fix clippy lints for Rust 1.97` — unblocks the workspace clippy job on current stable; no functional changes. Happy to split this into its own PR if you'd rather merge it independently.
2. `fix(network): align txmonitor codec with the on-the-wire protocol` — the existing local-tx-monitor codec diverged from the authoritative ouroboros-network codec in three ways that made interop with cardano-node and ogmios impossible:
   - `MsgAwaitAcquire` was guessed as wire label 4, which doesn't exist in the spec. The wire shares label 1 with `MsgAcquire`; peers disambiguate by protocol state (Idle = acquire, Acquired = await re-acquire). It now encodes as 1, and a stateless decode always yields `Acquire`, which agents map back based on their state.
   - The protocol-local `txmonitor::TxId` (the type used only by local-tx-monitor messages — not the chain-level transaction id types elsewhere in pallas) was modeled as a text string. The wire format is the hard-fork-combinator `GenTxId`: an era-wrapped `[era, hash-bytes]` pair. Clients such as ogmios treat the era wrapper as significant and probe the same hash under every plausible era.
   - `MsgGetMeasures` (11) / `MsgReplyGetMeasures` (12), part of the protocol since node-to-client v20, weren't modeled, so receiving them killed the decode.

   The client agent is repaired along the way: `release()` was rejected by its own outbound-state assertion, `Release`/`AwaitAcquire`/`Done` were missing from the assert tables, and `await_acquire()`, `query_measures()` and `done()` are added. The error type is renamed `Error` → `ClientError` (matching the chainsync `ClientError`/`ServerError` convention) to make room for the server agent.
3. `feat(network): add txmonitor server agent and wire it into NodeServer` — mirrors the localstate server pattern: `recv_while_idle` / `recv_while_acquired` drive the session and typed `send_*` replies cover NextTx, HasTx, GetSizes and GetMeasures. Wire label 1 received in the Acquired state surfaces as `ClientQueryRequest::AwaitAcquire`. `NodeServer` now subscribes `PROTOCOL_N2C_TX_MONITOR` and exposes the agent via `txmonitor()`.

## Why

Any pallas-based N2C server is currently unusable behind ogmios: ogmios opens local-tx-monitor eagerly on every session, and since accepting the N2C handshake claims support for the full protocol set, the unsubscribed protocol makes the multiplexer kill the whole connection. The immediate beneficiary is dolos, whose ouroboros N2C endpoint gains a local-tx-monitor session handler built on this agent — see the companion PR txpipe/dolos#1120.

## Breaking changes

- The protocol-local `txmonitor::TxId` changes shape (text string → era-wrapped `GenTxId`). No working code can regress on the wire format: the previous encoding was never accepted by real peers.
- `localtxmonitor` client error type renamed `Error` → `ClientError`.

## Testing

- Full client/server pair test over a unix socket exercising acquire, iteration, has-tx, sizes, measures, blocking re-acquire, release and termination.
- Interop-validated on mainnet with ogmios v7.0.0 as the client (through the dolos companion PR): mempool RPCs round-trip correctly and sessions survive the eager protocol-9 open. That deployment also fed a mainnet indexing pipeline whose derived state was verified byte-identical against a cardano-node reference, so the codec changes are exercised well beyond the unit suite.

One deliberate leniency: the server accepts `GetMeasures` regardless of the negotiated handshake version (the spec gates it at v20+). Pallas miniprotocol agents are version-agnostic by design, so the server errs on the lenient side.

## Summary by CodeRabbit

* **New Features**
  * Added transaction-monitor server support for node-to-client connections.
  * Added mempool measures querying, including transaction counts and size/capacity data.
  * Expanded transaction-monitor client and server interactions for snapshot management and transaction queries.
  * Improved support for newer transaction-monitor protocol behavior and message handling.

* **Tests**
  * Added integration coverage for transaction-monitor workflows, including acquisition, transaction queries, measures, release, and completion.

### DIFF
--- a/pallas-hardano/src/display/haskell_display.rs
+++ b/pallas-hardano/src/display/haskell_display.rs
@@ -681,8 +681,8 @@ impl HaskellDisplay for ValidityInterval {
     fn to_haskell_str(&self) -> String {
         format!(
             "(ValidityInterval {{invalidBefore = {}, invalidHereafter = {}}})",
-            &self.invalid_before.as_slot_no(),
-            &self.invalid_hereafter.as_slot_no()
+            self.invalid_before.as_slot_no(),
+            self.invalid_hereafter.as_slot_no()
         )
     }
 }
--- a/pallas-math/src/math.rs
+++ b/pallas-math/src/math.rs
@@ -441,8 +441,8 @@ mod tests {
                 (&one - &threshold_b).to_string(),
                 expected_threshold_b,
                 "(1 - f) *** b failed to match! - (1 - f)={}, b={}",
-                &c,
-                &b
+                c,
+                b
             );
 
             // do Taylor approximation for
@@ -467,14 +467,14 @@ mod tests {
             if a < threshold && res.estimation != ExpOrdering::LT {
                 panic!(
                     "wrong result should be leader {} should be more like {}",
-                    &temp, threshold
+                    temp, threshold
                 );
             }
 
             if a >= threshold && res.estimation != ExpOrdering::GT {
                 panic!(
                     "wrong result should not be leader {} should be more like {}",
-                    &temp, threshold
+                    temp, threshold
                 );
             }
 
--- a/pallas-network/src/facades.rs
+++ b/pallas-network/src/facades.rs
@@ -559,6 +559,8 @@ pub struct NodeServer {
     pub statequery: localstate::Server,
     /// Local-tx-submission server.
     pub localtxsubmission: localtxsubmission::Server,
+    /// Local-tx-monitor server.
+    pub txmonitor: txmonitor::Server,
     accepted_address: Option<UnixSocketAddr>,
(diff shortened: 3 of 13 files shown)


scarmuega:
@Javieracost looks good, but we need to rebase and force push to this branch before the merge. Should I do it?

Javieracost:
> @Javieracost looks good, but we need to rebase and force push to this branch before the merge. Should I do it?

Sure, please go ahead. Thanks :saluting_face:

PR #805, opened 2026-09-16
fix(hardano,primitives,utxorpc): follow-ups from ScriptNOfK's signed threshold (#803), plus a stack-safety gap in map_native_script

npoxon:
## Summary

Follow-ups from review on #803 (merged), which typed `ScriptNOfK`'s threshold `i64` instead of `u32`, plus a stack-safety gap in `pallas-utxorpc` found live in production.

## Changes

- `pallas-hardano`: the threshold now renders through `to_haskell_str_p()` like every other constructor argument in this file, so a negative value gets the required parens (`MOfN (-1) (...)`) instead of invalid syntax (`MOfN -1 (...)`).
- `dijkstra`: doc comment and a round-trip test for the same field, pinning `i64::MIN`/`-1`/`0`/`i64::MAX` and that the wire bytes match `alonzo`'s.
- `pallas-utxorpc`:
  - Added the missing upper-bound test for the `u5c` clamp (only the lower bound was covered).
  - Deduplicated `map_native_script` into the shared macro body - the two per-version copies differed only in one variant's name (`ScriptPubkey` vs `ScriptPubkeyHash`), now isolated in a small per-version helper.
  - **Made `map_native_script` stack-safe.** #802 fixed the NativeScript CBOR codec but explicitly excluded recursive downstream visitors, and this mapping was exactly that - plain recursion, one call frame per nesting level. Confirmed live: a production service hit a stack overflow converting the same 5,383-deep script from block 5183974 (the block that motivated #802) into u5c's wire format. Rewritten as a heap-backed work list, the same shape `pallas-primitives`' `Clone` impl already uses for the source type. No change to the mapped output (existing snapshot tests pass unchanged, plus a new mixed-shape test); a new test reverts the fix and confirms it aborts with a stack overflow on a 128 KiB thread without it.

## Known remaining gap

`map_native_script` itself is now stack-safe, but the returned `u5c::NativeScript` is a plain generated prost type with no custom `Clone`/`PartialEq`/`encode`/drop, so those still recurse on this value same as before. Measured ceilings on a 2 MiB stack (tokio's default worker stack), release build: `clone` ~10,640, `encode` ~32,895, `eq`/`drop` ~43,828. The production script is 5,383 deep, so this PR does resolve that crash in a release build, with `clone` the tightest margin at about 2x headroom - but a debug/dev-profile build's `clone` (~1,661) and `encode` (~2,051) still abort at that depth. Also note: the message this now successfully builds is rejected by a default-configured `prost` consumer (100-frame recursion limit) - the `no-recursion-limit` feature is a prerequisite for the end-to-end path, not an optimisation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Haskell output for negative native-script thresholds.
  * Prevented stack overflows when processing deeply nested native scripts.
  * Ensured native-script thresholds are safely clamped to supported ranges, including negative and oversized values.
  * Preserved the structure, ordering, and contents of mixed native-script trees during conversion.

* **Tests**
  * Added coverage for signed threshold encoding across the full 64-bit range.
  * Added tests for negative formatting, threshold clamping, deeply nested scripts, and mixed-tree conversion.


npoxon:
Hey @scarmuega, thank you for reviewing/merging my previous PR. This is a follow-up to address review comments from the previous PR.

scarmuega (approved):

